PR #13, opened 2015-03-04
Fix arm and arm64 mapping

Tyilo:
Requires the new `gum_arm64_writer_put_add_reg_reg_reg` function from https://github.com/frida/frida-gum/pull/17


oleavr:
Awesome!

PR #18, opened 2015-04-24
Open the files in text mode, not binary mode

marc1006:
Fix for python3:
Traceback (most recent call last):
File "../../../../frida-core/src/generate-api.py", line 180, in <module>
output_deps_file.write("glib-2.0")
TypeError: 'str' does not support the buffer interface mode

Another fix would be output_deps_file.write("glib-2.0".decode("UTF-8")) for all strings.

### DIFF
--- a/src/generate-api.py
+++ b/src/generate-api.py
@@ -176,12 +176,12 @@ if __name__ == '__main__':
                 for klass in api_classes:
                     klass.sort_members()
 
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! Hmm, it would be good to keep generating these files with LF even on Windows (our other headers are CRLF-free). Perhaps just do a search-replace and call a per-file function that does the decode and write?

oleavr:
Nah, let's go for this solution for now. Thanks!

PR #25, opened 2015-05-27
Add dummy.cpp file to force frida-server to link as C++

s1341:


server/Makefile.am
```diff
@@ -1,11 +1,13 @@
...
 
 frida_server_SOURCES = \
+	dummy.cpp \
 	server.vala
 frida_server_CFLAGS = \
 	-w
 frida_server_LDFLAGS = \
-	$(FRIDA_LDFLAGS)
+	$(FRIDA_LDFLAGS) \
+	-static
```
oleavr: Is this needed? I haven't checked what this will do on the other platforms.

s1341: No. I forgot to take it off. committing a fix for that now.

PR #30, opened 2015-06-10
Fix qinjector/injected thread race condition

s1341:
Fix qinjector so that it waits until the injected thread sends some data
on the pipe before waiting for the EOF which indicates that the injected
thread is done.

### DIFF
--- a/src/qnx/qinjector.vala
+++ b/src/qnx/qinjector.vala
@@ -53,14 +53,16 @@ namespace Frida {
 				return false;
 			});
 			timeout_source.attach (main_context);
-			ssize_t size;
-			try {
-				size = yield fifo.read_async (buf, Priority.DEFAULT, cancellable);
-			} catch (IOError e) {
-				if (e is IOError.CANCELLED)
-					throw new Error.TIMED_OUT ("Unexpectedly timed out while waiting for FIFO to establish");
-				else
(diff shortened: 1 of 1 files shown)


oleavr:
Not sure I understand how this can work; isn't a return value of 0 used to signal EOF?

s1341:
Yes. But we should always expect to receive some bytes from the injected thread *before* the pipe is closed. 

The race condition is that the host reaches that read before the injected thread has opened the pipe for writing. According to http://www.qnx.com/developers/docs/6.3.0SP3/neutrino/lib_ref/r/read.html:
```
When attempting to read from an empty pipe or FIFO:

1. If no process has the pipe open for writing, read() returns 0 to indicate end-of-file.
```

In order to prevent this, I added the while so that we always expect some bytes to be written from the injected thread *before* the pipe is closed. 

So far as I can tell this should work properly, including the timeout case, as the timeout will cause an exception.

oleavr:
I get that, but we're no longer using a FIFO but a unix socket, so I'm a bit confused why this is happening? The state machinery [here](https://github.com/frida/frida-core/blob/2a4513eb4ef9925da66d01f51989639de2992ad0/lib/pipe/pipe-posix.c#L202-L277) is supposed to handle this on the first attempt to read or write.

oleavr:
Oh wait, the injector uses a plain FIFO, sorry! This is correct. Let's merge.

s1341:
Awesome!

PR #34, opened 2015-06-14
Support enumerating apps on OS X

Tyilo:
The `[query startQuery];` should be run from the main thread according to Apple's docs, however this code works for at least `frida-ps -a`. (Maybe in this case the code is already running in the main thread?)

I'm not sure what we should do when multiple apps have the same bundle identifier.

### DIFF
--- a/src/darwin/system-darwin.m
+++ b/src/darwin/system-darwin.m
@@ -47,7 +47,7 @@ frida_system_init (void)
 FridaHostApplicationInfo *
 frida_system_enumerate_applications (int * result_length)
 {
-#ifdef HAVE_IOS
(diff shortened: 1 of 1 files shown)


Tyilo:
Another thing to consider: do we only want to show apps located in `/Applications`, `$HOME/Applications` and `/System/Library/CoreServices`?

oleavr:
Rad stuff! Hmm, yeah multiple apps with the same bundle identifier: food for thought.

LGTM except for minor inconsistencies in style (scope brace placement, spacing before cast and comment style).

oleavr:
Timed out. Please reopen later if you're interested in addressing the remaining issues.

PR #35, opened 2015-06-15
Check to see that the target is still alive before deallocing

s1341:
Otherwise, trying to dealloc will fail with an assertion, killing frida-server.

### DIFF
--- a/src/qnx/qinjector-glue.c
+++ b/src/qnx/qinjector-glue.c
@@ -136,19 +136,9 @@ frida_injection_instance_free (FridaInjectionInstance * instance)
   if (instance->remote_payload != 0)
   {
     GError * error = NULL;
-    gint fd;
-    gchar as_path[PATH_MAX];
-    gboolean result;
 
-    /* check to see if the target process is still alive before trying
-     * to dealloc: */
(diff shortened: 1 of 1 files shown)


oleavr:
Which assertion is failing? This seems like a workaround for a bug elsewhere.

s1341:
If the target process dies unexpectedly, `frida_injection_instance_free` will be called to cleanup. It will then try to do a remote dealloc, but fail on an assertion (here: https://github.com/frida/frida-core/blob/master/src/qnx/qinjector-glue.c#L658) because the target process no longer exists.

All I'm doing with this patch is checking that the process is still alive before deallocing.

oleavr:
That makes sense, but the process may die unexpectedly after our check, and we'll crash regardless. That's why I think we should solve this by not asserting on return values that rely on whether the target is still there or not.

s1341:
Ok. That might be a biggish change. I'll look into it tomorrow.

oleavr:
:+1:

s1341:
Ok. I changed all the asserts on fd != -1 to return -1 instead. This makes the check in frida_injection_instance_free unnecessary, so it has been removed.

PR #59, opened 2016-01-12
Fix bug when hooking pthread_create before it has been properly resol…

s1341:
…ved by the target application.

The bug is that we end up hooking the dynamic linker stub, instead of pthread_create. We will then get spurious calls to our replacement pthread_create when the application tries to resolve ANY import. This will obviously blow up.

### DIFF
--- a/lib/agent/agent-glue.c
+++ b/lib/agent/agent-glue.c
@@ -502,7 +502,7 @@ frida_get_address_of_thread_create_func (void)
 #ifdef G_OS_WIN32
   return GUM_FUNCPTR_TO_POINTER (_beginthreadex);
 #else
-  return GUM_FUNCPTR_TO_POINTER (pthread_create);
+  return dlsym (-1, "pthread_create");
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


lib/agent/agent-glue.c
```diff
@@ -502,7 +502,7 @@ frida_get_address_of_thread_create_func (void)
 #ifdef G_OS_WIN32
   return GUM_FUNCPTR_TO_POINTER (_beginthreadex);
 #else
-  return GUM_FUNCPTR_TO_POINTER (pthread_create);
+  return dlsym (-1, "pthread_create");
```
oleavr: Isn't it better to use `RTLD_DEFAULT` instead of `-1`, or is this define missing on one of your targets?

s1341:
We need RTLD_NEXT. It requires that we define _GNU_SOURCE, so I preferred to just use -1. I can push a version with RTLD_NEXT and _GNU_SOURCE

oleavr:
Ahh yes that one is indeed better, agreed. Feel free to squash the commits.

oleavr:
Nice!

PR #66, opened 2016-02-04
Prevent libdl from unloading to work around TLS destructor bug

s1341:
Keep a handle to libdl.so.2 so that it is never unloaded even when
libfrida-agent.so unloads in a process that did not previously have
libdl loaded.

This is to prevent a bug which was encountered where libdl tries to
perform a TLS key destructor after the libdl code has been unloaded,
resulting in a segfualt at thread cleanup time.

### DIFF
--- a/lib/agent/agent-glue.c
+++ b/lib/agent/agent-glue.c
@@ -55,6 +55,10 @@ static void frida_agent_on_log_message (const gchar * log_domain, GLogLevelFlags
 
 static void frida_agent_auto_ignorer_shutdown (FridaAgentAutoIgnorer * self);
 
+#ifdef HAVE_LINUX
+static void frida_libdl_prevent_unload ();
+#endif
+
 void
 frida_agent_environment_init (void)
 {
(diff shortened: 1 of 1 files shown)


lib/agent/agent-glue.c
```diff
@@ -98,7 +102,18 @@ frida_agent_environment_init (void)
...
   gum_script_backend_get_type (); /* Warm up */
   frida_error_quark (); /* Initialize early so GDBus will pick it up */
+#ifdef HAVE_LINUX
+  frida_libdl_prevent_unload ();
+#endif
+}
+
+#ifdef HAVE_LINUX
+void
+frida_libdl_prevent_unload ()
```
oleavr: Should be `(void)` instead of `()`, which in C means "any number of arguments", unlike in C++ where it means 0 arguments. Confusing :)

lib/agent/agent-glue.c
```diff
@@ -98,7 +102,18 @@ frida_agent_environment_init (void)
   gum_init ();
   gum_script_backend_get_type (); /* Warm up */
   frida_error_quark (); /* Initialize early so GDBus will pick it up */
+#ifdef HAVE_LINUX
+  frida_libdl_prevent_unload ();
```
oleavr: I think the condition should be "Linux but not Android".

oleavr:
Sweet!

PR #68, opened 2016-02-05
Missing execute bit (fixes #67)

### LINKED ISSUE #67: Injection fails on Windows 10 Metro/Modern apps
Metro apps on Windows 10 have more security than regular programs, so the dll injection fails
A possible fix would be to put the dll in `system32` (http://blog.nektra.com/main/2013/02/25/injecting-a-dll-in-modernui-metro-app-win8/ )

\cc @riverar

riverar:
Resolves crashing during spin up in Windows Store/Universal app scenarios


oleavr:
w00t, awesome!

PR #73, opened 2016-03-04
[iOS][FixBug]: app may be killed by watchdog.

Proteas:
Issue: app crashed, during "Instrumenting functions..."
Reason: killed by watchdog
Fix: use SBSLaunchApplicationForDebugging to launch app.
Note: my local toolchain is broken, so the code is not tested. please review and test the code first.

### DIFF
--- a/src/darwin/frida-helper-glue.m
+++ b/src/darwin/frida-helper-glue.m
@@ -462,33 +462,30 @@ _frida_helper_service_do_launch (FridaHelperService * self, const gchar * identi
 {
   NSAutoreleasePool * pool;
   FridaSpringboardApi * api;
-  NSDictionary * params, * options;
+  NSString * bundleIDStr = nil;
+  NSURL * urlToOpen = nil;
   UInt32 res;
 
   pool = [[NSAutoreleasePool alloc] init];
 
   api = _frida_get_springboard_api ();
 
-  params = [NSDictionary dictionary];
-
-  options = [NSDictionary dictionaryWithObject:@YES forKey:api->SBSApplicationLaunchOptionUnlockDeviceKey];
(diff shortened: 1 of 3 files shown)


Proteas:
this method is not perfect, because some app may not allow to be attached by a debugger.
maybe should change the timeout value of watchdog, also need to test that the related api is not denied by the Sandbox.
code snippet:
`BOOL SetWatchDogTimeout(int seconds)
{
    CFMutableDictionaryRef watchDogSrvInfo = NULL;
    watchDogSrvInfo = IOServiceMatching("IOWatchDogTimer");
    if (!watchDogSrvInfo) {
        printf("SetWatchDogTimeout: fail to get service info\n");
        return NO;
    }
    
    io_service_t watchDogSrv = NULL;
    watchDogSrv = IOServiceGetMatchingService(kIOMasterPortDefault, watchDogSrvInfo);
    if (!watchDogSrvInfo) {
        printf("SetWatchDogTimeout: fail to get service\n");
        return NO;
    }
    
    CFNumberRef timeout = CFNumberCreate(kCFAllocatorDefault, kCFNumberIntType, &seconds);
    if (!timeout) {
        printf("SetWatchDogTimeout: fail to create number object\n");
        return NO;
    }
    
    kern_return_t result = KERN_SUCCESS;
    result = IORegistryEntrySetCFProperties(watchDogSrv, timeout);
    
    CFRelease(timeout);
    IOObjectRelease(watchDogSrv);
    
    return (result == KERN_SUCCESS);
}`

oleavr:
Thanks! This is a known issue that I've been putting off by just making the code faster, so we don't run into the timeout. But, this is something we need to fix. I thought I saw some code earlier where you disabled the timeout, was that a dead end?

PR #74, opened 2016-03-08
Handle zombie processes gracefully

s1341:
These are processes which appear in /proc, but which you cannot open a /proc/<pid>/as for


oleavr:
:+1:

PR #96, opened 2016-12-14
Improve injectors

oleavr:
- Avoid private APIs on i/macOS
- Fix stability issue on i/macOS
- Fix Mach port leak on i/macOS
- Revise the agent entrypoint API to allow agents to stay resident. The old API was officially:

```C
void
frida_agent_main (const char * data)
{
  /*
   * Agent will stay loaded until this function returns, after which Frida
   * will emit the `uinjected` signal.
   */
}
```

But it actually had two extra arguments for internal use:

```C
void
frida_agent_main (const char * data,
                  const GumMemoryRange * mapped_range,
                  GumThreadId parent_thread_id)
{
}
```

* `mapped_range`: was (and is still) used to tell [Stalker](https://github.com/frida/frida-gum/blob/master/gum/backend-x86/gumstalker-x86.c) to ignore our memory range when our shared library was manually mapped into the process. This means it is `NULL` if it was loaded by e.g. `dlopen()`, in which case frida-agent walks the loaded libraries to find the memory range.
* `parent_thread_id`: on i/macOS we used to have a Mach thread sticking around, waiting for our POSIX thread to stop. We needed to know its thread ID so we could exclude it from `Process.enumerateThreads()`. As of this PR, this thread no longer sticks around, so this argument is now gone.

The new API is officially:

```C
#include <stdbool.h>

void
frida_agent_main (const char * data,
                  bool * stay_resident)
{
  /*
   * Uncomment to prevent agent from getting unloaded after returning from this function.
   * This also means the `uninjected` signal is not emitted, and Frida will stop keeping track
   * of this injected thread.
   */
  // *stay_resident = true;
}
```

But internally it is actually:

```C
#include <stdbool.h>

void
frida_agent_main (const char * data,
                  bool * stay_resident,
                  const GumMemoryRange * mapped_range)
{
  *stay_resident = true;
}
```

Though I think we should consider the third argument public now, as it's useful for some applications, e.g. using Stalker.

@s1341: I didn't dare to port the Linux/MIPS and QNX/ARM portions as I'm not able to test them, but as those are passing `NULL, NULL` for the last two arguments, it just means that `stay_resident` is unavailable for those ports (for now).


s1341:
@oleavr I don't currently have access to test devices for linux-arm, linux-mips or qnx-arm. I suggest we leave those as is for the moment.

oleavr:
@s1341 Sounds like a good plan.

PR #116, opened 2017-03-25
Fix deadlock caused by breaking inside libSystem initialization

mrmacete:
Place a second hardware breakpoint at the end of dylb's `ImageLoader::recursiveInitialization` to let it release the spinlock and notify the end of libSystem initialization. This is done reusing the live disassembly tool. dyld is found in memory like before, going back from main thread's instruction pointer.

### DIFF
--- a/src/darwin/frida-helper-backend-glue.m
+++ b/src/darwin/frida-helper-backend-glue.m
@@ -32,6 +32,7 @@
 
 typedef struct _FridaHelperContext FridaHelperContext;
 typedef struct _FridaSpawnInstance FridaSpawnInstance;
+typedef guint FridaBreakpointPhase;
 typedef struct _FridaInjectInstance FridaInjectInstance;
 typedef struct _FridaInjectPayloadLayout FridaInjectPayloadLayout;
 typedef struct _FridaAgentDetails FridaAgentDetails;
@@ -40,15 +41,6 @@ typedef struct _FridaAgentEmitContext FridaAgentEmitContext;
 
 typedef struct _FridaExceptionPortSet FridaExceptionPortSet;
 typedef union _FridaDebugState FridaDebugState;
-
-typedef guint FridaBreakpointPhase;
-
-enum _FridaBreakpointPhase
-{
-  FRIDA_BREAKPOINT_FIRST,
-  FRIDA_BREAKPOINT_SECOND,
-  FRIDA_BREAKPOINT_DONE
-};
 
 struct _FridaHelperContext
 {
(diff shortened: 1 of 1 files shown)


src/darwin/frida-helper-backend-glue.m
```diff
@@ -217,9 +227,12 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
 static kern_return_t frida_set_debug_state (mach_port_t thread, gconstpointer state, GumCpuType cpu_type);
 static void frida_set_hardware_breakpoint (gpointer state, GumAddress break_at, GumCpuType cpu_type);
 
+static void frida_breakpoint_receive_request (FridaSpawnInstance * self);
+static void frida_breakpoint_send_response (FridaSpawnInstance * self);
```
oleavr: Since these guys are behaving like member functions of `Frida.SpawnInstance`, I'd say they should be moved to right before `on_server_recv()`, and renamed to something like:

```C
static void frida_spawn_instance_receive_breakpoint_request (FridaSpawnInstance * self);
static void frida_spawn_instance_send_breakpoint_response (FridaSpawnInstance * self);
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -852,12 +870,18 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+   *   won't be finished yet.
    * - Swap out the thread's exception ports with our own.
    * - Resume the task.
    * - Wait until we get a message on our exception port, meaning our breakpoint
    *   was hit.
+   * - Place a second hardware breakpoint at the end of dyld's
+   *   ImageLoader::recursiveInitialization method, in order to let it release
+   *   the spin lock and notify the complete initialization of libSystem
+   * - Send a response to the message, resuming the thread
+   * - Wait until we get a second message on our exception port
```
oleavr: These new bullets should end with full stop for consistency.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -2456,6 +2535,43 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
 #endif
 }
 
+static void
+frida_breakpoint_receive_request (FridaSpawnInstance * self)
+{
+  __Request__exception_raise_state_identity_t * request = &self->pending_request;
+  mach_msg_header_t * header;
+  kern_return_t ret;
```
oleavr: Suggest a blank line after this one.

PR #119, opened 2017-04-01
Rewrite early instrumentation using thread hijacking

mrmacete:
- Get hold of the main thread to read its instruction pointer, which will
  tell us where dyld is in memory.
- Walk backwards to find dyld's Mach-O header.
- Walk its symbols and find a function that's called at a point where the
  process is sufficiently initialized to load frida-agent, but early enough
  so that app's initializer still didn't run. In this case we choose
  `dyld::initializeMainExecutable()`. At the beginning of this function dyld is
  initialized but libSystem is still missing.
- Set a hardware breakpoint at the beginning of this function.
- Swap out the thread's exception ports with our own.
- Resume the task.
- Wait until we get a message on our exception port, meaning our breakpoint
  was hit.
- Hijack thread's instruction pointer to call `dlopen("/usr/lib/libSystem.B.dylib")`
  and then return back to the beginning of `initializeMainExecutable()` and restore
  previous thread state.
- Swap back the thread's orginal exception ports.
- Clear the hardware breakpoint by restoring the thread's debug registers.

It's actually more complex than that, because:
- This doesn't work on newer versions of dyld because to call `dlopen()` it's
  necessary to `registerThreadHelpers()` first, which is normally done by libSystem
  itself during its initialization.
- To overcome this catch-22 we alloc a fake `LibSystemHelpers` object and register
  it (also by hijacking thread's instruction pointer like descripted above).
- On older dyld versions, registering helpers before loading libSystem led to
  crashes, so we detect this condition and unset the helpers before calling `dlopen()`,
  by writing a `NULL` directly into the global `dyld::gLibSystemHelpers` because in
  some dyld versions calling `registerThreadHelpers(NULL)` caused a `NULL` dereference.
- At the end of `dlopen()`, we set the global `libSystemInitialized` flag present in
  the global `dyld::qProcessInfo` structure, because on newer dyld versions that doesn't
  happen automatically due to the presence of our fake helpers.
- One of the functions provided by the helper should return a buffer for the errors,
  but since our fake helpers object implements its functions only using a return,
  it will not return any buffer. To avoid this to happen, we set a breakpoint also
  on `dyld:dlerrorClear()` function and inject an immediate return,
  effectively disabling the function.
- At the end of `dlopen()` we finally deallocate our fake helpers (because now they've
  been replaced by real libSystem ones) and the string we used as a parameter for dlopen.

### DIFF
--- a/src/darwin/frida-helper-backend-glue.m
+++ b/src/darwin/frida-helper-backend-glue.m
@@ -32,6 +32,7 @@
 
 typedef struct _FridaHelperContext FridaHelperContext;
 typedef struct _FridaSpawnInstance FridaSpawnInstance;
+typedef struct _FridaSpawnInstanceDyldData FridaSpawnInstanceDyldData;
 typedef guint FridaBreakpointPhase;
 typedef struct _FridaInjectInstance FridaInjectInstance;
 typedef struct _FridaInjectPayloadLayout FridaInjectPayloadLayout;
@@ -73,7 +74,9 @@ struct _FridaSpawnInstance
   GumCpuType cpu_type;
   mach_port_t thread;
   FridaDebugState previous_debug_state;
-  FridaDebugState second_debug_state;
+  FridaDebugState dlerror_clear_debug_state;
+  FridaDebugState ret_gadget_debug_state;
+  FridaDebugState breakpoint_debug_state;
 
   mach_port_t server_port;
   dispatch_source_t server_recv_source;
@@ -82,13 +85,32 @@ struct _FridaSpawnInstance
   __Request__exception_raise_state_identity_t pending_request;
 
   FridaBreakpointPhase breakpoint_phase;
+  mach_vm_address_t lib_name;
+  mach_vm_address_t fake_helpers;
+  mach_vm_address_t dyld_data;
+  GumAddress dlopen_address;
+  GumAddress info_address;
+  GumAddress register_helpers_address;
+  GumAddress helpers_ptr_address;
+  GumAddress ret_gadget;
+  mach_port_t task;
+  GumDarwinUnifiedThreadState previous_thread_state;
 };
 
 enum _FridaBreakpointPhase
 {
-  FRIDA_BREAKPOINT_FIRST,
-  FRIDA_BREAKPOINT_SECOND,
+  FRIDA_BREAKPOINT_SET_HELPERS,
+  FRIDA_BREAKPOINT_DLOPEN_LIBC,
+  FRIDA_BREAKPOINT_SKIP_CLEAR,
+  FRIDA_BREAKPOINT_JUST_RETURN,
+  FRIDA_BREAKPOINT_CLEANUP,
   FRIDA_BREAKPOINT_DONE
+};
+
+struct _FridaSpawnInstanceDyldData
+{
+  const char libc[32];
+  guint8 helpers[32];
 };
 
 struct _FridaInjectInstance
@@ -203,7 +225,13 @@ static void frida_spawn_instance_resume (FridaSpawnInstance * self);
 
 static void frida_spawn_instance_receive_breakpoint_request (FridaSpawnInstance * self);
 static void frida_spawn_instance_send_breakpoint_response (FridaSpawnInstance * self);
+static void frida_spawn_instance_create_dyld_data (FridaSpawnInstance * self);
+static void frida_spawn_instance_destroy_dyld_data (FridaSpawnInstance * self);
+static void frida_spawn_instance_unset_helpers (FridaSpawnInstance * self);
+static gboolean frida_spawn_instance_is_libc_initialized (FridaSpawnInstance * self);
 static void frida_spawn_instance_on_server_recv (void * context);
+static void frida_spawn_instance_call_set_helpers (FridaSpawnInstance * self, GumDarwinUnifiedThreadState state, mach_vm_address_t helpers);
+static void frida_spawn_instance_call_dlopen (FridaSpawnInstance * self, GumDarwinUnifiedThreadState state, mach_vm_address_t lib_name, int mode);
 
 static void frida_make_pipe (int fds[2]);
 
@@ -228,11 +256,8 @@ static kern_return_t frida_get_debug_state (mach_port_t thread, gpointer state, 
 static kern_return_t frida_set_debug_state (mach_port_t thread, gconstpointer state, GumCpuType cpu_type);
 static void frida_set_hardware_breakpoint (gpointer state, GumAddress break_at, GumCpuType cpu_type);
 
-static gboolean frida_store_base_address_if_libc (const GumModuleDetails * details, gpointer user_data);
-static GumAddress frida_find_libc_initializer (guint task, GumAddress base);
 static GumAddress frida_find_function_end (guint task, GumCpuType cpu_type, GumAddress start, gsize max_size);
 static csh frida_create_capstone (GumCpuType cpu_type, GumAddress start);
-static GumAddress frida_get_module_slide (gconstpointer command, gsize ncmds, GumAddress base);
 
 static void frida_mapper_library_blob_deallocate (FridaMappedLibraryBlob * self);
 
@@ -847,13 +872,10 @@ _frida_darwin_helper_backend_prepare_spawn_instance_for_injection (FridaDarwinHe
   mach_msg_type_number_t state_count = GUM_DARWIN_THREAD_STATE_COUNT;
   thread_state_flavor_t state_flavor = GUM_DARWIN_THREAD_STATE_FLAVOR;
   GumAddress dyld_start, dyld_granularity, dyld_chunk, dyld_header;
-  GumAddress libc_header, probe_address, recursive_init_address, second_probe_address;
+  GumAddress probe_address, dlerror_clear_address;
   GumDarwinModule * dyld;
-  FridaDebugState breakpoint_debug_state;
   FridaExceptionPortSet * previous_ports;
   dispatch_source_t source;
-  static GumAddress cached_address_for_breakpoint[6] = { 0, };
-  static GumAddress cached_ret_offset[6] = { 0, };
 
   /*
    * We POSIX_SPAWN_START_SUSPENDED which means that the kernel will create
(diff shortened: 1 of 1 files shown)


src/darwin/frida-helper-backend-glue.m
```diff
@@ -82,15 +85,34 @@
...
   FridaBreakpointPhase breakpoint_phase;
+  mach_vm_address_t lib_name;
+  mach_vm_address_t fake_helpers;
+  mach_vm_address_t dyld_data;
+  GumAddress dlopen_address;
+  GumAddress info_address;
+  GumAddress register_helpers_address;
+  GumAddress helpers_ptr_address;
+  GumAddress ret_gadget;
+  uint task;
```
oleavr: Guess this one should be `mach_port_t`?

src/darwin/frida-helper-backend-glue.m
```diff
@@ -862,27 +884,46 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+   *   previous thread state.
    * - Swap back the thread's orginal exception ports.
    * - Clear the hardware breakpoint by restoring the thread's debug registers.
    *
+   * It's actually more complex than that, because:
+   * - This doesn't work on newer versions of dyld because to call dlopen() it's
+   *   necessary to registerThreadHelpers() first, which is normally done by libSystem
+   *   itself during its initialization.
+   * - To overcome this catch-22 we alloc a fake LibSystemHelpers object and register
+   *   it (also by hijacking thread's instruction pointer like descripted above).
```
oleavr: Tiny typo: `described`, and a tiny tweak while at it: `as described above`.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -862,27 +884,46 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+   * - On older dyld versions, registering helpers before loading libSystem led to
+   *   crashes, so we detect this condition and unset the helpers before calling dlopen(),
+   *   by writing a NULL directly into the global dyld::gLibSystemHelpers because in
+   *   some dyld version calling registerThreadHelpers(NULL) caused a NULL dereference.
```
oleavr: Couple of tweaks:
versions*
causes*

src/darwin/frida-helper-backend-glue.m
```diff
@@ -936,70 +978,47 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
-    if (second_probe_address == 0)
-      goto handle_probe_dyld_error;
-
-    cached_ret_offset[instance->cpu_type] = second_probe_address - recursive_init_address;
-  }
-  else
-  {
-    second_probe_address = recursive_init_address + cached_ret_offset[instance->cpu_type];
+    instance->dlopen_address |= 1;
+    instance->register_helpers_address |= 1;
```
oleavr: This seems like a bug in `GumDarwinModule`, but we can ignore that for now.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1556,26 +1567,284 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
 }
 
 static void
+frida_spawn_instance_create_dyld_data (FridaSpawnInstance * self)
+{
+  kern_return_t ret;
+  gboolean bret;
```
oleavr: Maybe `write_succeeded` is clearer. And changing the assertion to: `g_assert (write_succeeded);` so it's slightly easier to debug.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1556,26 +1567,284 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+  FridaSpawnInstanceDyldData data = { "/usr/lib/libSystem.B.dylib", {0,}};
```
oleavr: A few tiny tweaks needed:
```
  FridaSpawnInstanceDyldData data = { "/usr/lib/libSystem.B.dylib", { 0, } };
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1556,26 +1567,284 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+  ret = vm_deallocate (self->task, (vm_address_t) self->dyld_data, sizeof (FridaSpawnInstanceDyldData));
+  g_assert_cmpint (ret, ==, KERN_SUCCESS);
+
+  self->dyld_data = 0;
+}
+
+static void
+frida_spawn_instance_unset_helpers (FridaSpawnInstance * self)
+{
+  gboolean bret;
```
oleavr: (Like above.)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1556,26 +1567,284 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+  }
+
+  g_assert_cmpint (bret, ==, TRUE);
+}
+
+static void
+frida_spawn_instance_set_libc_initialized (FridaSpawnInstance * self)
+{
+  GumAddress initialized_address;
+  gboolean bret;
```
oleavr: (Like above.)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1609,6 +1885,124 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+    state.uts.ts64.__rsi = mode;
+
+    temp = current_pc;
+    state.uts.ts64.__rsp -= 16;
+    bret = gum_darwin_write (self->task, state.uts.ts64.__rsp, &temp, sizeof (temp));
+    g_assert_cmpint (bret, ==, TRUE);
+  }
+  else
+  {
+    uint32 temp [3];
```
oleavr: Extra space.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1609,6 +1885,124 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+frida_spawn_instance_call_set_helpers (FridaSpawnInstance * self, GumDarwinUnifiedThreadState state, mach_vm_address_t helpers)
+{
+  kern_return_t ret;
+  GumAddress current_pc;
+
+#ifdef HAVE_I386
+  if (self->cpu_type == GUM_CPU_AMD64)
+  {
+    uint64 temp;
+    gboolean bret;
```
oleavr: (Like above.)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1609,6 +1885,124 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+
+    temp = current_pc;
+    state.uts.ts64.__rsp -= 8;
+    bret = gum_darwin_write (self->task, state.uts.ts64.__rsp, &temp, sizeof (temp));
+    g_assert_cmpint (bret, ==, TRUE);
+  }
+  else
+  {
+    uint32 temp [2];
+    gboolean bret;
```
oleavr: (Like above.)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1609,6 +1885,124 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+frida_spawn_instance_call_dlopen (FridaSpawnInstance * self, GumDarwinUnifiedThreadState state, mach_vm_address_t lib_name, int mode)
+{
+  kern_return_t ret;
+  GumAddress current_pc;
+
+#ifdef HAVE_I386
+  if (self->cpu_type == GUM_CPU_AMD64)
+  {
+    uint64 temp;
+    gboolean bret;
```
oleavr: (Like above.)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1609,6 +1885,124 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+    state.uts.ts64.__rdi = helpers;
+
+    temp = current_pc;
+    state.uts.ts64.__rsp -= 8;
+    bret = gum_darwin_write (self->task, state.uts.ts64.__rsp, &temp, sizeof (temp));
+    g_assert_cmpint (bret, ==, TRUE);
+  }
+  else
+  {
+    uint32 temp [2];
```
oleavr: Extra space.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1609,6 +1885,124 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+
+    temp = current_pc;
+    state.uts.ts64.__rsp -= 16;
+    bret = gum_darwin_write (self->task, state.uts.ts64.__rsp, &temp, sizeof (temp));
+    g_assert_cmpint (bret, ==, TRUE);
+  }
+  else
+  {
+    uint32 temp [3];
+    gboolean bret;
```
oleavr: (Like above.)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1609,6 +1885,124 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+  {
+    uint64 temp;
+    gboolean bret;
+
+    current_pc = state.uts.ts64.__rip;
+    state.uts.ts64.__rip = self->dlopen_address;
+    state.uts.ts64.__rdi = lib_name;
+    state.uts.ts64.__rsi = mode;
+
+    temp = current_pc;
```
oleavr: This can be dropped as `GumAddress` is the same size.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1556,26 +1567,284 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+      }
+#else
+      if (self->cpu_type == GUM_CPU_ARM64)
+      {
+        state.ts_64.__pc = state.ts_64.__lr;
+      }
+      else
+      {
+        state.ts_32.__pc = state.ts_32.__lr;
+      }
```
oleavr: Blocks should be omitted here and above.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1556,26 +1567,284 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+
+      break;
+
+    case FRIDA_BREAKPOINT_DLOPEN_LIBC:
+      if (frida_spawn_instance_is_libc_initialized (self))
+        frida_spawn_instance_unset_helpers (self);
+
+      memcpy (&state, &self->previous_thread_state, sizeof (state));
+
+      frida_spawn_instance_call_dlopen (self, state, self->lib_name, RTLD_LAZY | RTLD_GLOBAL);
```
oleavr: Nit: we usually specify these two flags in this order: `RTLD_GLOBAL | RTLD_LAZY`.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1609,6 +1885,124 @@ static gboolean frida_agent_context_init_functions (FridaAgentContext * self, Gu
...
+  if (self->cpu_type == GUM_CPU_AMD64)
+  {
+    uint64 temp;
+    gboolean bret;
+
+    current_pc = state.uts.ts64.__rip;
+    state.uts.ts64.__rip = self->register_helpers_address;
+    state.uts.ts64.__rdi = helpers;
+
+    temp = current_pc;
```
oleavr: This can be dropped as GumAddress is the same size.

mrmacete:
pushed --force to fix typos in commit message :blush:

oleavr:
Thanks! :)

PR #126, opened 2017-05-13
Add FRIDA_GADGET_LISTEN_ADDRESS environment variable to frida-gadget

esanfelix:
This change introduces FRIDA_GADGET_LISTEN_ADDRESS environment variable. If the variable is set, frida-gadget will use it as a listen address. Otherwise the default localhost:27042 is used.

For example, the following makes it listen on any interface:

```
FRIDA_GADGET_LISTEN_ADDRESS="tcp:host=192.168.178.11,port=27042" DYLD_INSERT_LIBRARIES=./build/frida-macos-universal/lib/FridaGadget.dylib ./build/frida-macos-universal/bin/frida-server
```

And then:
```
➜  Desktop netstat -ant | grep 27042
tcp4       0      0  192.168.178.11.27042   *.*                    LISTEN
➜  Desktop
```

This allows using frida-gadget with remote systems (e.g. a linux armhf system) that do not have "USB" support without doing any network/vm-level forwarding.

### DIFF
--- a/lib/gadget/gadget.vala
+++ b/lib/gadget/gadget.vala
@@ -1,4 +1,7 @@
 namespace Frida.Gadget {
+	private const string DEFAULT_LISTEN_ADDRESS = "127.0.0.1";
+	private const uint16 DEFAULT_LISTEN_PORT = 27042;
+
 	private enum State {
 		CREATED,
 		STARTED,
@@ -17,9 +20,6 @@ namespace Frida.Gadget {
 	private Gum.Exceptor exceptor;
 	private Mutex mutex;
 	private Cond cond;
-
-	private const string DEFAULT_LISTEN_ADDRESS = "127.0.0.1";
-	private const uint16 DEFAULT_LISTEN_PORT = 27042;
 
 	public void load () {
 		if (loaded)
@@ -73,27 +73,6 @@ namespace Frida.Gadget {
 		mutex.unlock ();
 	}
 
-	private bool try_get_listen_address (out string listen_address, out uint16 listen_port, out string listen_uri) {
-		try {
-			var env_listen_address = GLib.Environment.get_variable ("FRIDA_GADGET_LISTEN_ADDRESS");
-			var raw_address = (env_listen_address != null) ? env_listen_address : DEFAULT_LISTEN_ADDRESS;
-			var socket_address = NetworkAddress.parse (raw_address, DEFAULT_LISTEN_PORT).enumerate ().next ();
-			if (socket_address is InetSocketAddress) {
-				var inet_socket_address = socket_address as InetSocketAddress;
-				var inet_address = inet_socket_address.get_address ();
-				var family = (inet_address.get_family () == SocketFamily.IPV6) ? "ipv6" : "ipv4";
-				listen_address = inet_address.to_string ();
-				listen_port = inet_socket_address.get_port ();
-				listen_uri = "tcp:family=%s,host=%s,port=%hu".printf (family, listen_address, listen_port);
-				return true;
-			} else {
-				return false;
-			}
-		} catch (GLib.Error e) {
-			return false;
-		}
-	}
-
 	private async void start () {
 		var gadget_range = memory_range ();
 		Gum.Cloak.add_range (gadget_range);
@@ -125,7 +104,6 @@ namespace Frida.Gadget {
 			}
 			resume ();
 		} else {
-
 			string listen_address;
 			uint16 listen_port;
 			string listen_uri;
@@ -142,6 +120,31 @@ namespace Frida.Gadget {
 			} catch (GLib.Error e) {
 				log_error ("Failed to start: " + e.message);
 			}
+		}
+	}
+
+	private bool try_get_listen_address (out string listen_address, out uint16 listen_port, out string listen_uri) {
+		listen_address = null;
+		listen_port = 0;
+		listen_uri = null;
+
+		try {
+			var env_listen_address = GLib.Environment.get_variable ("FRIDA_GADGET_LISTEN_ADDRESS");
+			var raw_address = (env_listen_address != null) ? env_listen_address : DEFAULT_LISTEN_ADDRESS;
+			var socket_address = NetworkAddress.parse (raw_address, DEFAULT_LISTEN_PORT).enumerate ().next ();
+			
+			if (!(socket_address is InetSocketAddress))
+				return false;
+
+			var inet_socket_address = socket_address as InetSocketAddress;
+			var inet_address = inet_socket_address.get_address ();
+			var family = (inet_address.get_family () == SocketFamily.IPV6) ? "ipv6" : "ipv4";
+			listen_address = inet_address.to_string ();
+			listen_port = inet_socket_address.get_port ();
+			listen_uri = "tcp:family=%s,host=%s,port=%hu".printf (family, listen_address, listen_port);
+			return true;
+		} catch (GLib.Error e) {
+			return false;
 		}
 	}
 
@@ -418,7 +421,6 @@ namespace Frida.Gadget {
 	}
 
 	private class Server : Object {
-
 		private unowned Gum.ScriptBackend script_backend = null;
 		private string listen_uri;
 		private bool jit_enabled;
--- a/lib/gadget/gadget.vala
+++ b/lib/gadget/gadget.vala
@@ -73,6 +73,27 @@ namespace Frida.Gadget {
 		mutex.unlock ();
 	}
 
+	private bool try_get_listen_address (out string listen_address, out uint16 listen_port, out string listen_uri) {
+		try {
+			var env_listen_address = GLib.Environment.get_variable ("FRIDA_GADGET_LISTEN_ADDRESS");
+			var raw_address = (env_listen_address != null) ? env_listen_address : DEFAULT_LISTEN_ADDRESS;
+			var socket_address = NetworkAddress.parse (raw_address, DEFAULT_LISTEN_PORT).enumerate ().next ();
+			if (socket_address is InetSocketAddress) {
+				var inet_socket_address = socket_address as InetSocketAddress;
+				var inet_address = inet_socket_address.get_address ();
+				var family = (inet_address.get_family () == SocketFamily.IPV6) ? "ipv6" : "ipv4";
+				listen_address = inet_address.to_string ();
+				listen_port = inet_socket_address.get_port ();
+				listen_uri = "tcp:family=%s,host=%s,port=%hu".printf (family, listen_address, listen_port);
+				return true;
+			} else {
+				return false;
+			}
+		} catch (GLib.Error e) {
+			return false;
+		}
+	}
+
 	private async void start () {
 		var gadget_range = memory_range ();
 		Gum.Cloak.add_range (gadget_range);
@@ -105,28 +126,19 @@ namespace Frida.Gadget {
 			resume ();
 		} else {
 
+			string listen_address;
+			uint16 listen_port;
 			string listen_uri;
+			if (!try_get_listen_address (out listen_address, out listen_port, out listen_uri)) {
+				log_error ("Invalid listen address");
+				return;
+			}
+
 			try {
-				string? listen_address = GLib.Environment.get_variable ("FRIDA_GADGET_LISTEN_ADDRESS");
-				var raw_address = (listen_address != null) ? listen_address : DEFAULT_LISTEN_ADDRESS;
-				var socket_address = NetworkAddress.parse (raw_address, DEFAULT_LISTEN_PORT).enumerate ().next ();
-				if (socket_address is InetSocketAddress) {
-					var inet_socket_address = socket_address as InetSocketAddress;
-					var inet_address = inet_socket_address.get_address ();
-					var family = (inet_address.get_family () == SocketFamily.IPV6) ? "ipv6" : "ipv4";
-					string address = inet_address.to_string ();
-					uint16 port = inet_socket_address.get_port ();
-					listen_uri = "tcp:family=%s,host=%s,port=%hu".printf (family, address, port);
-
 					var s = new Server (listen_uri, jit_enabled, gadget_range);
-
 					yield s.start ();
 					server = s;
-					log_info ("Listening on " + address + " TCP port " + port.to_string ());
-
-				} else {
-					log_error ("Invalid listen address\n");
-				}
+					log_info ("Listening on " + listen_address + " TCP port " + listen_port.to_string());
 			} catch (GLib.Error e) {
 				log_error ("Failed to start: " + e.message);
 			}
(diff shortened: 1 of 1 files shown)


lib/gadget/gadget.vala
```diff
@@ -398,8 +398,12 @@ namespace Frida.Gadget {
 		}
 
 		public async void start () throws Error {
+			string listen_address = GLib.Environment.get_variable ("FRIDA_GADGET_LISTEN_ADDRESS");
```
oleavr: - This should be done at the higher level, and passed into the constructor instead – ordered first, just before `jit_enabled`. It could be a `string?`, and if `null` is passed in it will use `LISTEN_ADDRESS` instead.
- We should parse the address (and port if specified) instead of expecting the user to provide a cryptic URI string. Feel free to steal the code from `frida-server`: https://github.com/frida/frida-core/blob/3b3bc6c4779eca14f3c9cdf48c45f28f5b3944d5/server/server.vala#L39-L49

lib/gadget/gadget.vala
```diff
@@ -398,8 +398,12 @@ namespace Frida.Gadget {
...
+			if (listen_address == null){
```
oleavr: Missing space after `)`.

esanfelix: I agree that'd be better.

Would you prefer the parsing done only before starting the server, i.e. after https://github.com/frida/frida-core/blob/master/lib/gadget/gadget.vala#L102 , or parse it anyway?

esanfelix: What would you say of this:

```
--- a/lib/gadget/gadget.vala
+++ b/lib/gadget/gadget.vala
@@ -18,6 +18,9 @@ namespace Frida.Gadget {
        private Mutex mutex;
        private Cond cond;

+       private const string DEFAULT_LISTEN_ADDRESS = "127.0.0.1";
+       private const uint16 DEFAULT_LISTEN_PORT = 27042;
+
        public void load () {
                if (loaded)
                        return;
@@ -90,6 +93,7 @@ namespace Frida.Gadget {

                var script_file = GLib.Environment.get_variable ("FRIDA_GADGET_SCRIPT");
                bool jit_enabled = GLib.Environment.get_variable ("FRIDA_GADGET_ENABLE_JIT") != null;
+
                if (script_file != null) {
                        var r = new ScriptRunner (script_file, jit_enabled, gadget_range);
                        try {
@@ -100,12 +104,30 @@ namespace Frida.Gadget {
                        }
                        resume ();
                } else {
-                       var s = new Server (jit_enabled, gadget_range);
+
+                       string listen_uri;
                        try {
-                               yield s.start ();
-                               server = s;
-                               log_info ("Listening on TCP port 27042");
-                       } catch (Error e) {
+                               string? listen_address = GLib.Environment.get_variable ("FRIDA_GADGET_LISTEN_ADDRESS");
+                               var raw_address = (listen_address != null) ? listen_address : DEFAULT_LISTEN_ADDRESS;
+                               var socket_address = NetworkAddress.parse (raw_address, DEFAULT_LISTEN_PORT).enumerate ().next ();
+                               if (socket_address is InetSocketAddress) {
+                                       var inet_socket_address = socket_address as InetSocketAddress;
+                                       var inet_address = inet_socket_address.get_address ();
+                                       var family = (inet_address.get_family () == SocketFamily.IPV6) ? "ipv6" : "ipv4";
+                                       string address = inet_address.to_string ();
+                                       uint16 port = inet_socket_address.get_port ();
+                                       listen_uri = "tcp:family=%s,host=%s,port=%hu".printf (family, address, port);
+
+                                       var s = new Server (listen_uri, jit_enabled, gadget_range);
+
+                                       yield s.start ();
+                                       server = s;
+                                       log_info ("Listening on " + address + " TCP port " + port.to_string ());
+
+                               } else {
+                                       log_error ("Invalid listen address\n");
+                               }
+                       } catch (GLib.Error e) {
                                log_error ("Failed to start: " + e.message);
                        }
                }
@@ -384,22 +406,23 @@ namespace Frida.Gadget {
        }

        private class Server : Object {
-               private const string LISTEN_ADDRESS = "tcp:host=127.0.0.1,port=27042";

                private unowned Gum.ScriptBackend script_backend = null;
                private bool jit_enabled;
                private Gum.MemoryRange gadget_range;
                private DBusServer server;
                private Gee.HashMap<DBusConnection, Client> clients = new Gee.HashMap<DBusConnection, Client> ();
+               private string listen_uri;

-               public Server (bool jit_enabled, Gum.MemoryRange gadget_range) {
+               public Server (string listen_uri, bool jit_enabled, Gum.MemoryRange gadget_range) {
                        this.jit_enabled = jit_enabled;
                        this.gadget_range = gadget_range;
+                       this.listen_uri = listen_uri;
                }

                public async void start () throws Error {
                        try {
-                               server = new DBusServer.sync (LISTEN_ADDRESS, DBusServerFlags.AUTHENTICATION_ALLOW_ANONYMOUS, DBus.generate_guid ());
+                               server = new DBusServer.sync (listen_uri, DBusServerFlags.AUTHENTICATION_ALLOW_ANONYMOUS, DBus.generate_guid ());
                        } catch (GLib.Error listen_error) {
                                throw new Error.ADDRESS_IN_USE (listen_error.message);
                        }
```

lib/gadget/gadget.vala
```diff
@@ -384,22 +406,22 @@ namespace Frida.Gadget {
...
 		private Gum.MemoryRange gadget_range;
 		private DBusServer server;
 		private Gee.HashMap<DBusConnection, Client> clients = new Gee.HashMap<DBusConnection, Client> ();
+		private string listen_uri;
 
-		public Server (bool jit_enabled, Gum.MemoryRange gadget_range) {
+		public Server (string listen_uri, bool jit_enabled, Gum.MemoryRange gadget_range) {
 			this.jit_enabled = jit_enabled;
 			this.gadget_range = gadget_range;
+			this.listen_uri = listen_uri;
```
oleavr: Assignments should follow the same order as their declarations.

lib/gadget/gadget.vala
```diff
@@ -384,22 +406,22 @@ namespace Frida.Gadget {
...
 
 	private class Server : Object {
-		private const string LISTEN_ADDRESS = "tcp:host=127.0.0.1,port=27042";
-
 		private unowned Gum.ScriptBackend script_backend = null;
 		private bool jit_enabled;
 		private Gum.MemoryRange gadget_range;
 		private DBusServer server;
 		private Gee.HashMap<DBusConnection, Client> clients = new Gee.HashMap<DBusConnection, Client> ();
+		private string listen_uri;
```
oleavr: Declaration order should follow the constructor, so right before `jit_enabled`.

lib/gadget/gadget.vala
```diff
@@ -100,12 +104,30 @@ namespace Frida.Gadget {
...
+					listen_uri = "tcp:family=%s,host=%s,port=%hu".printf (family, address, port);
+
+					var s = new Server (listen_uri, jit_enabled, gadget_range);
+
+					yield s.start ();
+					server = s;
+					log_info ("Listening on " + address + " TCP port " + port.to_string ());
+
+				} else {
+					log_error ("Invalid listen address\n");
```
oleavr: Newline will be added automatically, so the `\n` should be omitted.

lib/gadget/gadget.vala
```diff
@@ -100,12 +104,30 @@ namespace Frida.Gadget {
...
+				string? listen_address = GLib.Environment.get_variable ("FRIDA_GADGET_LISTEN_ADDRESS");
+				var raw_address = (listen_address != null) ? listen_address : DEFAULT_LISTEN_ADDRESS;
+				var socket_address = NetworkAddress.parse (raw_address, DEFAULT_LISTEN_PORT).enumerate ().next ();
+				if (socket_address is InetSocketAddress) {
+					var inet_socket_address = socket_address as InetSocketAddress;
+					var inet_address = inet_socket_address.get_address ();
+					var family = (inet_address.get_family () == SocketFamily.IPV6) ? "ipv6" : "ipv4";
+					string address = inet_address.to_string ();
+					uint16 port = inet_socket_address.get_port ();
+					listen_uri = "tcp:family=%s,host=%s,port=%hu".printf (family, address, port);
```
oleavr: This feels like it belongs in a separate function, or at least in a separate try/catch block above. But I'd lean towards a function, perhaps something like:

```Vala
string listen_address;
uint16 listen_port;
string listen_uri;
if (!try_get_listen_address (out listen_address, out listen_port, out listen_uri)) {
	log_error ("Invalid listen address");
	return;
}

try {
	var s = new Server (listen_uri, jit_enabled, gadget_range);
	yield s.start ();

	server = s;

	log_info ("Listening on %s TCP port %hu", listen_address, listen_port);
} catch (GLib.Error e) {
	log_error ("Failed to start: %s", e.message);
}
```

and then:

```Vala
private bool try_get_listen_address (out string address, out uint16 port, out string uri) {
}
```

esanfelix:
I think I covered it all. I get some warnings on usage of unassigned variables,  but I believe they should always be assigned if try_get_listen_address() returns true.

Are you fine with that, or do you suggest adding any checks or anything else? Note these are the first few lines of vala code I ever touch, so I might very well not be doing things in the best possible way.

lib/gadget/gadget.vala
```diff
@@ -384,22 +418,23 @@ namespace Frida.Gadget {
 	}
 
 	private class Server : Object {
-		private const string LISTEN_ADDRESS = "tcp:host=127.0.0.1,port=27042";
```
oleavr: This blank line should also be removed.

lib/gadget/gadget.vala
```diff
@@ -100,12 +125,21 @@ namespace Frida.Gadget {
...
+
 			try {
-				yield s.start ();
-				server = s;
-				log_info ("Listening on TCP port 27042");
-			} catch (Error e) {
+					var s = new Server (listen_uri, jit_enabled, gadget_range);
+					yield s.start ();
+					server = s;
+					log_info ("Listening on " + listen_address + " TCP port " + listen_port.to_string());
```
oleavr: Using a format string would be a bit more readable, as `log_info()` supports it natively.

lib/gadget/gadget.vala
```diff
@@ -100,12 +125,21 @@ namespace Frida.Gadget {
 			}
 			resume ();
 		} else {
-			var s = new Server (jit_enabled, gadget_range);
+
```
oleavr: Let's remove this blank line.

lib/gadget/gadget.vala
```diff
@@ -70,6 +73,27 @@ namespace Frida.Gadget {
 		mutex.unlock ();
 	}
 
+	private bool try_get_listen_address (out string listen_address, out uint16 listen_port, out string listen_uri) {
+		try {
+			var env_listen_address = GLib.Environment.get_variable ("FRIDA_GADGET_LISTEN_ADDRESS");
+			var raw_address = (env_listen_address != null) ? env_listen_address : DEFAULT_LISTEN_ADDRESS;
+			var socket_address = NetworkAddress.parse (raw_address, DEFAULT_LISTEN_PORT).enumerate ().next ();
+			if (socket_address is InetSocketAddress) {
```
oleavr: Early return here would be better (less nesting).

lib/gadget/gadget.vala
```diff
@@ -70,6 +73,27 @@ namespace Frida.Gadget {
 		mutex.unlock ();
 	}
 
+	private bool try_get_listen_address (out string listen_address, out uint16 listen_port, out string listen_uri) {
```
oleavr: This method is lower level than `start()`, so it should be below it. Perhaps right below it, so it's easy to visually spot all the environment variables being accessed.

lib/gadget/gadget.vala
```diff
@@ -18,6 +18,9 @@ namespace Frida.Gadget {
 	private Mutex mutex;
 	private Cond cond;
 
+	private const string DEFAULT_LISTEN_ADDRESS = "127.0.0.1";
+	private const uint16 DEFAULT_LISTEN_PORT = 27042;
+
```
oleavr: Constants should be first, followed by a blank line (like you did here).

oleavr:
> I get some warnings on usage of unassigned variables, but I believe they 
> should always be assigned if try_get_listen_address() returns true.

Just assign them to `null` and `0` to make the compiler happy. (We should keep the code warning-free, even if this is a case where the compiler's static analysis isn't adequate.)

oleavr:
Well done btw! 👍

lib/gadget/gadget.vala
esanfelix: First meaning even before the enum's ?

oleavr: Yep I'd say so.

lib/gadget/gadget.vala
esanfelix: Does it? I tried at first, but I get:

```
gadget.vala:140.6-140.87: error: 2 extra arguments for `void Frida.Gadget.log_info (string message)'
					log_info ("Listening on %s TCP port %hu", listen_address, listen_port.to_string());
					^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

esanfelix: I didn't quite find log_info, but frida_gadget_log_info() calls g_info("%s", message); so I'd say log_info does not support format strings?

Besides this one, the only requested changes should be there I think.

oleavr: Oops, I forgot that this was a gadget-internal function. In that case just do: `log_info ("foo: %s bar: %d".printf (foo, bar));`

lib/gadget/gadget.vala
```diff
@@ -100,17 +104,50 @@ namespace Frida.Gadget {
...
+	private bool try_get_listen_address (out string listen_address, out uint16 listen_port, out string listen_uri) {
+		listen_address = null;
+		listen_port = 0;
+		listen_uri = null;
+
+		try {
+			var env_listen_address = GLib.Environment.get_variable ("FRIDA_GADGET_LISTEN_ADDRESS");
+			var raw_address = (env_listen_address != null) ? env_listen_address : DEFAULT_LISTEN_ADDRESS;
+			var socket_address = NetworkAddress.parse (raw_address, DEFAULT_LISTEN_PORT).enumerate ().next ();
+
```
oleavr: Trailing whitespace.

lib/gadget/gadget.vala
esanfelix: Right, I could have seen that option in the code but missed it :)

esanfelix:
Btw is there any doc on coding guidelines so we can perhaps save some review rounds in future PRs?

Like where to put/remove white-spaces, ordering of variables in constructors and declarations should match, etc etc.

oleavr:
> Btw is there any doc on coding guidelines so we can perhaps save some review rounds in future PRs?
>
> Like where to put/remove white-spaces, ordering of variables in constructors and declarations should match, etc etc.

There is none I'm afraid, but I've been meaning to write that up, just haven't gotten around to it yet.

PR #145, opened 2017-10-01
On MacOS, create a hidden subdirectory to use for tmp

wallneradam:
This way the trash icon can remain empty, so you don't want to always empty it.


oleavr:
Good idea, thanks!

PR #150, opened 2017-10-10
XNU Agent reload.sh and readme improvements

stek29:


lib/xnu/README.md
```diff
@@ -8,9 +8,9 @@ Experimental kernel driver for boosting Frida's capabilities on macOS:
 
 ## Installation
 
-1. Open `FridaKernelAgent.xcodeproj` and build it with Xcode
-2. Review `reload.sh`, adjusting it to match your system
-3. Run `reload.sh` to install and load the driver
+Run `reload.sh` to install and load the driver
+If it fails to load, set `CODE_SIGN_IDENTITY` enviroment variable to
+identity valid for Kext signing, or disable SIP
```
oleavr: Sweet! How about reflowing to 72 cols, with full stops between sentences? (Which in hindsight I should have done in my list.)

lib/xnu/reload.sh
```diff
@@ -1,7 +1,23 @@
...
-sudo chown -R root /Library/Extensions/FridaKernelAgent.kext
-sudo chgrp -R wheel /Library/Extensions/FridaKernelAgent.kext
-sudo kextload /Library/Extensions/FridaKernelAgent.kext
+
+KEXT_NAME=FridaKernelAgent.kext
+
+sudo kextunload "/Library/Extensions/$KEXT_NAME" &>/dev/null
+sudo rm -rf "/Library/Extensions/$KEXT_NAME"
+
+BUILD_DIR="$(mktemp -d)"
```
oleavr: Guess we should clean this up afterwards?

lib/xnu/reload.sh
```diff
@@ -1,7 +1,23 @@
...
+
+cd "$(dirname "$(realpath "$0")")"
+xcodebuild build\
+  CONFIGURATION_BUILD_DIR="$BUILD_DIR"\
+  CODE_SIGN_IDENTITY="$CODE_SIGN_IDENTITY"\
+  &>/dev/null
```
oleavr: For consistency it would be good to put a space before each continuation-backslash.

lib/xnu/reload.sh
stek29: but it's a temp dir...

lib/xnu/reload.sh
stek29: I forgot sh allows it :|

lib/xnu/reload.sh
oleavr: I know, just don't like leaving stuff there. (OCD :) )

oleavr:
```

Comment (oleavr) [reply lib/xnu/reload.sh]: Perhaps adding this right after would do the trick:
```
function cleanup {
  rm -rf "$BUILD_DIR"
}
trap cleanup EXIT
```

PR #162, opened 2018-01-12
fixes the frida agent injection bug on android 8 arm 32bit processes

madushan1000:
This fixed the arm 32bit processes crashing on android 8(O) after the last release.


oleavr:
Thanks!

PR #167, opened 2018-02-14
support bytecode loading for gadget

0xh0b0:
Detects script type by looking for duktape bytecode magic value
closes frida/frida#427


oleavr:
Thanks a lot! I've merged this fix with two tiny style tweaks. Releasing 10.6.53 right now with this change.

PR #169, opened 2018-02-15
FIX: access array out of range

0xh0b0:


oleavr:
Oops, I also missed this. :-) Thanks!

PR #177, opened 2018-03-23
Allow mapped_range to be passed on gadget load()

mrmacete:
- for proper cloaking
- implemented for i/macOS
- as a special entry in apple[] array
- if that’s not present, then fallback to usual behavior

### DIFF
--- a/lib/gadget/gadget-glue.c
+++ b/lib/gadget/gadget-glue.c
@@ -72,13 +72,7 @@ static void on_keep_alive_timer_fire (CFRunLoopTimerRef timer, void * info);
 static FridaFoundationApi * frida_foundation_api_try_get (void);
 static FridaCFApi * frida_cf_api_try_get (void);
 static FridaObjCApi * frida_objc_api_try_get (void);
-static gboolean frida_dylib_range_try_get (const char * apple[], GumMemoryRange * range);
-
-# if GLIB_SIZEOF_VOID_P == 4
-#  define GUM_PFMT "%llx"
-# else
-#  define GUM_PFMT "%lx"
-# endif
+static gboolean frida_dylib_range_try_get (const gchar * apple[], GumMemoryRange * range);
 
 #endif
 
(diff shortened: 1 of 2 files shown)


lib/gadget/gadget-glue.c
```diff
@@ -585,4 +609,22 @@ frida_objc_api_try_get (void)
   return api;
 }
 
+static gboolean
+frida_dylib_range_try_get (const char * apple[], GumMemoryRange * range)
+{
+  const char * entry;
+  int i = 0;
```
oleavr: Nit: for consistency it would be good to use `gchar` and `gint` here, and perhaps use `guint` for the index while at it.

lib/gadget/gadget-glue.c
```diff
@@ -585,4 +609,22 @@ frida_objc_api_try_get (void)
...
+
+  while ((entry = apple[i++]) != NULL)
+  {
+    if (!strncmp (entry, "frida_dylib_range=", 18))
```
oleavr: To make this a little bit clearer, how about:

```C
    if (g_str_has_prefix (entry, "frida_dylib_range="))
```

lib/gadget/gadget-glue.c
```diff
@@ -585,4 +609,22 @@ frida_objc_api_try_get (void)
...
+    {
+      if (sscanf (entry, "frida_dylib_range=0x" GUM_PFMT ",0x%lx", &range->base_address, &range->size) == 2)
```
oleavr: Using `G_GINT64_MODIFIER` and `G_GSIZE_MODIFIER` here would be more precise and portable.

lib/gadget/gadget-glue.c
```diff
@@ -72,6 +72,13 @@ static void on_keep_alive_timer_fire (CFRunLoopTimerRef timer, void * info);
 static FridaFoundationApi * frida_foundation_api_try_get (void);
 static FridaCFApi * frida_cf_api_try_get (void);
 static FridaObjCApi * frida_objc_api_try_get (void);
+static gboolean frida_dylib_range_try_get (const char * apple[], GumMemoryRange * range);
```
oleavr: Nit: would be good to use `gchar` here.

PR #184, opened 2018-04-23
Keep reference to client alive (WIP)

mrmacete:
This is needed because latest vala compiler doesn't keep references automatically.

### DIFF
--- a/server/server.vala
+++ b/server/server.vala
@@ -390,15 +390,16 @@ namespace Frida.Server {
 				connection.unregister_object (registration_id);
 			}
 
-			private void schedule_idle (owned ScheduledFunc function) {
+			private void schedule_idle (owned ScheduledFunc func) {
 				var client = this;
-				Idle.add(() => {
-					function (client);
+				Idle.add (() => {
+					func ();
+					client = null;
 					return false;
 				});
 			}
 
-			private delegate void ScheduledFunc (Client client);
+			private delegate void ScheduledFunc ();
 
 			private GLib.DBusMessage on_connection_message (DBusConnection connection, owned DBusMessage message, bool incoming) {
 				DBusMessage result = message;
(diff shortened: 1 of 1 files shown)


server/server.vala
```diff
@@ -448,8 +448,9 @@ namespace Frida.Server {
 					path.scanf ("/re/frida/AgentSession/%u", out session_id);
 					if (member == "Close") {
 						if (type != DBusMessageType.METHOD_CALL) {
+                            var client = this;
 							Idle.add (() => {
```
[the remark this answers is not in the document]
mrmacete: you mean something like this? : 

```
public void schedule_idle (owned SourceFunc function) {
	Idle.add((owned) function);
}
```

and then:

```
	schedule_idle (() => {
		sessions.remove (session_id);
		return false;
	});
```

but that alone isn't enough to keep a reference to `this`... it would be great do do something inside `schedule_idle` to take care of that

server/server.vala
```diff
@@ -390,6 +390,16 @@ namespace Frida.Server {
 				connection.unregister_object (registration_id);
 			}
 
+			private void schedule_idle (owned ScheduledFunc function) {
+				var client = this;
+				Idle.add(() => {
```
oleavr: Missing space before `(`.

server/server.vala
```diff
@@ -390,6 +390,16 @@ namespace Frida.Server {
 				connection.unregister_object (registration_id);
 			}
 
+			private void schedule_idle (owned ScheduledFunc function) {
```
oleavr: Nit: perhaps we should name the variable `func` to match the delegate name. In general I try to make variable names no more verbose than their corresponding types, so either we should rename `ScheduledFunc` to `ScheduledFunction`, or make `function` just `func`. As GLib uses `FooFunc` in a few places I tend to stick to that.

server/server.vala
```diff
@@ -390,6 +390,16 @@ namespace Frida.Server {
...
+				var client = this;
+				Idle.add(() => {
+					function (client);
```
oleavr: How about dropping the argument from the delegate and instead doing:

```Vala
client = null;
```

after calling `func`?

PR #189, opened 2018-06-04
Improved Windows childgating by attaching to CreateProcessInternal

ineedblood:
`CreateProcessA/W`and `CreateProcessInternalA` end all all calling `CreateProcessInternalW` in the end. 

The flow should be something such as:

`CreateProcessA` > `CreateProcessInternalA` > `CreateProcessInternalW`
`CreateProcessW` > `CreateProcessInternalW`

Therefore, we would avoid double hooks with just watching CreateProcessInternalW

### DIFF
--- a/lib/agent/agent.vala
+++ b/lib/agent/agent.vala
@@ -1189,7 +1189,7 @@ namespace Frida.Agent {
 			var interceptor = Gum.Interceptor.obtain ();
 
 #if WINDOWS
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks a lot! Committed as fe815b3 with one small tweak: argument indices were off by one due to `CreateProcessInternalW()`'s extra argument at index 0, so arguments weren't parsed correctly and the process was not created suspended.

Just tagged Frida 11.0.7 with these changes – binaries should be available shortly.

Cheers!

PR #194, opened 2018-07-13
Fix i/macOS libc++ initialization on early instrumentation

mrmacete:
Fix libSystem initialization when DYLD_INSERT_LIBRARIES is there

In this case, `dlopen()` takes a different code path and hits this check in `ImageLoaderMachO::doModInitFunctions()`:

```
if ( ! dyld::gProcessInfo->libSystemInitialized ) {
  const char* installPath = getInstallPath();
  if ( (installPath == NULL) || (strcmp(installPath, LIBSYSTEM_DYLIB_PATH) != 0) )
    dyld::throwf("initializer in image (%s) that does not link with libSystem.dylib\n", this->getPath());
}
```

When we call it to initialize libSystem it throws an exception when loading `libc++.dylib` which is a dependency but it’s not libSystem itself, leaving c++ not initialized and leading to potential crashes.

This solution boils down to hook dyld’s `strcmp()` and make it successful in any case when called from the context of the above code.

note: on macOS the issue was there even without using DYLD_INSERT_LIBRARY

### DIFF
--- a/src/darwin/frida-helper-backend-glue.m
+++ b/src/darwin/frida-helper-backend-glue.m
@@ -2505,7 +2505,8 @@ frida_spawn_instance_on_server_recv (void * context)
   }
 
 #ifndef HAVE_I386
-  if (pc == (self->ret_gadget & ~1)) {
+  if (pc == (self->ret_gadget & ~1))
+  {
     if (self->cpu_type == GUM_CPU_ARM64)
       state.ts_64.__pc = state.ts_64.__lr;
     else
@@ -4104,7 +4105,6 @@ frida_find_next_instruction (mach_port_t task, GumCpuType cpu_type, GumAddress s
 
   return address;
 }
-
 
 static csh
 frida_create_capstone (GumCpuType cpu_type, GumAddress start)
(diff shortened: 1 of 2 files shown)


src/darwin/frida-helper-backend-glue.m
```diff
@@ -2438,39 +2472,57 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+    memcpy (&self->previous_thread_state, &state, sizeof (state));
+
     if (pc == self->modern_entry_address)
       self->breakpoint_phase = FRIDA_BREAKPOINT_CLEANUP;
     else
       self->breakpoint_phase = self->need_helpers ? FRIDA_BREAKPOINT_SET_HELPERS : FRIDA_BREAKPOINT_DLOPEN_LIBC;
   }
 
+#ifndef HAVE_I386
+  if (pc == (self->ret_gadget & ~1)) {
```
oleavr: Brace on the wrong line.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -3940,6 +4079,33 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+  cs_disasm_iter (capstone, &code, &size, &address, insn);
+
+  cs_free (insn, 1);
+  cs_close (&capstone);
+  g_free (image);
+
+  return address;
+}
+
+
```
oleavr: Extra newline.

PR #195, opened 2018-07-18
Fix some issues with uclibc and MIPS support

r0ck3tAKATrashPanda:
This fixes the MIPS ASM syntax errors, as well as changing the syntax for struct field access using uclibc. I am not sure if the latter is a overarching issue, but I had issues while trying to compile.


oleavr:
Thanks!

PR #197, opened 2018-07-20
Fix MIPS regressions

r0ck3tAKATrashPanda:
This is to complete [#195](https://github.com/frida/frida-core/pull/195). I missed a register correction.


oleavr:
Thanks!

Issue #193, opened 2018-07-12
[Electra] Unable to spawn apps using dyld3


AeonLucid:
Same issue as #176, but for newer versions.

**Device:** iPhone 6
**iOS version:** 11.3.1
**Jailbreak:** Electra-mptcp
**Frida tools & server**: 12.0.3

(1) Commands:
- `frida-ps -U -f com.apple.DocumentsApp` (Stock Files app)
- `frida-ps -U -f com.toyopagroup.picaboo` (Snapchat)

Result:
```
Failed to spawn: unexpectedly timed out while waiting for process to suspend
```

(2) Commands after the above error while keeping the spawned app open:
- `frida-ps -U -n Files`
- `frida-ps -U -n Snapchat`

Result:
```
Failed to attach: unexpected error while probing dyld of target process
```

_If you open the app manually, it attaches fine at (2)._

xnamurt:
I've the same issue

ChiChou:
👍

mrmacete:
it's because hardware breakpoints don't work from iOS 11.3.1 up to and including iOS 11.4.1

we currently rely on hardware breakpoints for spawn and early instrumentation on i/macOS

currently working on implementing software breakpoints for those iOS versions

bihihinok:
@oleavr what about android , please check issue #575 , #576

PR #199, opened 2018-07-26
Refactor i/macOS early instrumentation

mrmacete:
- simplify the logic, by having a single interface for setting breakpoints and a single place to set the thread and debug states
- implement software breakpoints on iOS versions from 11.3.1 up to and including 11.4.1 where hardware breakpoints are broken in the kernel
- use hardware single step to get out of repeatable breakpoint on all platform, saving a call to capstone
- implement the `getThreadBufferFor_dlerror` helpers function using breakpoints,  needed for iOS 11.3.1

Tested on iOS 9.0.2, 9.3.3, 10.2, 11.1.2, 11.3.1.

According to my tests, the performance of software breakpoints seems about the same as hardware breakpoints.

this should fix https://github.com/frida/frida-core/issues/193


AeonLucid:
Thanks!

PR #215, opened 2018-10-24
Adding support for systemless root

tobypetrice:
A lot of rooted Android devices are doing so called "systemless root". One particular example is the Pixel root by Chainfire. In it, the system partition is moved to "system_root", so searching for things like "/system/bin/linker" fails. 

This patch lets Frida work on systemless rooted Android devices.


oleavr:
Wow, great catch! Thanks a lot for this fix, and sorry for the delay here.

PR #217, opened 2018-10-30
Add retry logic around task_info call

mrmacete:
Apparently task_info fails randomly with “invalid argument” (and sometimes “mig server died”) because we’re calling it too early in task’s lifetime. It will now retry at most 3 times before giving up.


oleavr:
Awesome! :metal: Thanks!

PR #219, opened 2018-11-08
Avoid deadlock when a thread suspends itself

mrmacete:
By calling `thread_suspend` right away without getting locks.

### DIFF
--- a/lib/agent/agent.vala
+++ b/lib/agent/agent.vala
@@ -1720,8 +1720,8 @@ namespace Frida.Agent {
 			if (Gum.Cloak.has_thread (thread_id))
 				return 0;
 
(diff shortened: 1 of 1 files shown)


lib/agent/agent.vala
```diff
@@ -1720,8 +1720,9 @@ namespace Frida.Agent {
 			if (Gum.Cloak.has_thread (thread_id))
 				return 0;
 
+			uint caller_thread_id = (uint) Gum.Process.get_current_thread_id ();
 			var script_backend = Runner.shared_instance.script_backend;
```
oleavr:
```suggestion
			var script_backend = Runner.shared_instance.script_backend;
			uint caller_thread_id = (uint) Gum.Process.get_current_thread_id ();
```

PR #220, opened 2018-11-15
Add frida-server options to specify the temp subdirectory

psychomario:
Tested on x86_64 linux

Option `-d` allows for changing the frida-server temporary sub-directory, into which `frida-helper-{32,64}` are dropped, relative to the temporary directory, falling back on the original directory (`re.frida.server`)

```
I   ➤​  ./build/frida-linux-x86_64/bin/frida-server -d FRIDA_TEST
```

Results in:

```
I   ➤​  ls /tmp/FRIDA_TEST                                                                                                                                                       [~/caps/frida]
frida-helper-32*  frida-helper-64*
```

### DIFF
--- a/server/server.vala
+++ b/server/server.vala
@@ -3,8 +3,10 @@ namespace Frida.Server {
 
 	private const string DEFAULT_LISTEN_ADDRESS = "127.0.0.1";
 	private const uint16 DEFAULT_LISTEN_PORT = 27042;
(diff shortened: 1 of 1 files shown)


psychomario:
Apologies, I did not understand how the listen host specification worked. I have removed the commits that change this code and updated the above accordingly.

oleavr:
Thanks! No worries – being able to pick the directory is a nice feature though!

PR #221, opened 2018-12-13
Steal task port in _inject

mrmacete:
To avoid it to be destroyed too early in case the app takes too long to be loaded.


mrmacete:
force-pushed with the new attempt to fix this

mrmacete:
right :joy:! i'm blind

PR #222, opened 2019-01-11
frida-inject: read script from stdin if dash is specified

iGio90:
Scenario: i recently bought an android device which needs to do something through injector.
This device for some reason that i don't want to know doesn't allow "setenforce 0" on the stock kernel.
The enforcing selinux doesn't allow frida injector to read files from outside it's context.
An higher level solution wich would bring benefit in other situations would be providing the script source straight to the injector through other ways.
Discussed with the owner both, a base64'd input to avoid issues with \n or an input to the stdin (through outputstream).
I'm gonna use this through an android app which uses Process to run a su -c command to invoke the injector. Tested and working by providing the script to the outputstream
and later by closing it (eof).

### DIFF
--- a/inject/inject.vala
+++ b/inject/inject.vala
@@ -54,7 +54,7 @@ namespace Frida.Inject {
 			printerr ("Path to JavaScript file must be specified\n");
 			return 3;
 		} else if (script_path == "-") {
-		    script_path = null;
+			script_path = null;
 			script_source = read_stdin ();
 		}
 
--- a/inject/inject.vala
+++ b/inject/inject.vala
@@ -242,7 +242,7 @@ namespace Frida.Inject {
 		public async void start () throws Error {
 			yield load ();
 
-			if (enable_development) {
+			if (enable_development && script_path != null) {
 				try {
 					script_monitor = File.new_for_path (script_path).monitor_file (FileMonitorFlags.NONE);
 					script_monitor.changed.connect (on_script_file_changed);
@@ -287,7 +287,7 @@ namespace Frida.Inject {
 			try {
 				string name;
 				string source;
-				if (script_source == null) {
+				if (script_path != null) {
 					name = Path.get_basename (script_path).split (".", 2)[0];
 					try {
 						FileUtils.get_contents (script_path, out source);
--- a/inject/inject.vala
+++ b/inject/inject.vala
@@ -48,12 +48,13 @@ namespace Frida.Inject {
 			return 2;
 		}
 
-		string? script_source = null;
-
 		if (script_path == null || script_path == "") {
 			printerr ("Path to JavaScript file must be specified\n");
 			return 3;
-		} else if (script_path == "-") {
+		}
+
+		string? script_source = null;
+		if (script_path == "-") {
 			script_path = null;
 			script_source = read_stdin ();
 		}
@@ -286,7 +287,6 @@ namespace Frida.Inject {
 			try {
 				string name;
 				string source;
-
 				if (script_source == null) {
 					name = Path.get_basename (script_path).split (".", 2)[0];
 					try {
--- a/inject/inject.vala
+++ b/inject/inject.vala
@@ -4,6 +4,7 @@ namespace Frida.Inject {
 	private static int target_pid = -1;
 	private static string? target_name;
 	private static string script_path;
+	private static string script_source;
 	private static bool eternalize;
 	private static bool enable_jit;
 	private static bool enable_development;
@@ -20,6 +21,21 @@ namespace Frida.Inject {
 		{ null }
 	};
 
+	private static string read_stdin () {
+		var input = new StringBuilder ();
+		while (!stdin.eof ()) {
+			var buffer = new char[1024];
+			string read_chunk = stdin.gets (buffer);
+			if (read_chunk != null) {
+				input.append (read_chunk);
+			} else {
+				break;
+			}
+		}
+		return input.str;
+	}
+
+
 	private static int main (string[] args) {
 #if !WINDOWS
 		Posix.setsid ();
@@ -28,6 +44,7 @@ namespace Frida.Inject {
 		Environment.init ();
 
 		script_path = "";
+		script_source = "";
 
 		try {
 			var ctx = new OptionContext ();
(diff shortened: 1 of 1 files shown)


inject/inject.vala
```diff
@@ -20,6 +21,21 @@ namespace Frida.Inject {
...
+	private static string read_stdin () {
+		var input = new StringBuilder ();
+		while (!stdin.eof ()) {
+			var buffer = new char[1024];
+			string read_chunk = stdin.gets (buffer);
+			if (read_chunk != null) {
+				input.append (read_chunk);
+			} else {
+				break;
+			}
```
oleavr: I'd simplify this to just:

```js
			if (read_chunk == null)
				break;
			input.append (read_chunk);
```

inject/inject.vala
```diff
@@ -20,6 +21,21 @@ namespace Frida.Inject {
 		{ null }
 	};
 
+	private static string read_stdin () {
```
oleavr: Should be moved below `main()`, so the code can be read like a newspaper – top to bottom.

inject/inject.vala
```diff
@@ -53,9 +70,11 @@ namespace Frida.Inject {
 		if (script_path == "") {
 			printerr ("Path to JavaScript file must be specified\n");
 			return 3;
+		} else if (script_path == "-") {
+			script_source = read_stdin();
```
oleavr: Missing space before `(`.

inject/inject.vala
```diff
@@ -261,13 +288,18 @@ namespace Frida.Inject {
...
 
 			try {
-				var name = Path.get_basename (script_path).split (".", 2)[0];
-
 				string source;
-				try {
-					FileUtils.get_contents (script_path, out source);
-				} catch (FileError e) {
-					throw new Error.INVALID_ARGUMENT (e.message);
+				string name;
```
oleavr: Ordering of these two should be swapped. We first set `name`, then `source`.

inject/inject.vala
```diff
@@ -197,6 +222,7 @@ namespace Frida.Inject {
 	private class ScriptRunner : Object {
 		private Script script;
 		private string script_path;
+		private string script_source;
```
oleavr: These should both be nullable, i.e. `string?`.

inject/inject.vala
```diff
@@ -206,9 +232,10 @@ namespace Frida.Inject {
 		private Gee.HashMap<string, PendingResponse> pending = new Gee.HashMap<string, PendingResponse> ();
 		private int64 next_request_id = 1;
 
-		public ScriptRunner (Session session, string script_path, bool enable_jit, bool enable_development) {
+		public ScriptRunner (Session session, string script_path, string script_source, bool enable_jit, bool enable_development) {
```
oleavr: These should both be nullable, i.e. `string?`.

inject/inject.vala
```diff
@@ -96,6 +115,11 @@ namespace Frida.Inject {
 			construct;
 		}
 
+		public string script_source {
+			get;
+			construct;
+		}
```
oleavr: These should both be nullable, i.e. `string?`.

inject/inject.vala
```diff
@@ -4,6 +4,7 @@ namespace Frida.Inject {
 	private static int target_pid = -1;
 	private static string? target_name;
 	private static string script_path;
+	private static string script_source;
```
oleavr: This should be a local variable.

inject/inject.vala
```diff
@@ -53,9 +70,11 @@ namespace Frida.Inject {
 		if (script_path == "") {
```
oleavr: Suggest adding the `script_source` local variable before this line:

```js
		string? script_source = null;
```

inject/inject.vala
```diff
@@ -4,6 +4,7 @@ namespace Frida.Inject {
 	private static int target_pid = -1;
 	private static string? target_name;
 	private static string script_path;
```
oleavr: This should now be nullable.

inject/inject.vala
```diff
@@ -28,6 +44,7 @@ namespace Frida.Inject {
 		Environment.init ();
 
 		script_path = "";
+		script_source = "";
```
oleavr: These two lines should be dropped.

inject/inject.vala
oleavr:
```suggestion
		if (script_path == null || script_path == "") {
```

inject/inject.vala
```diff
@@ -53,9 +70,11 @@ namespace Frida.Inject {
 		if (script_path == "") {
 			printerr ("Path to JavaScript file must be specified\n");
 			return 3;
+		} else if (script_path == "-") {
```
oleavr: Should also set `script_path` to `null` inside this block – let's do that as the first statement.

inject/inject.vala
```diff
@@ -20,6 +21,21 @@ namespace Frida.Inject {
 		{ null }
 	};
 
+	private static string read_stdin () {
+		var input = new StringBuilder ();
+		while (!stdin.eof ()) {
+			var buffer = new char[1024];
```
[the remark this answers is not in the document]
iGio90: ok, so it will keep use the allocated buffer and rewrite it? on java you need to allocate a new one! nice to know!

inject/inject.vala
```diff
@@ -50,12 +48,17 @@ namespace Frida.Inject {
...
-		if (script_path == "") {
+		string? script_source = null;
+
+		if (script_path == null || script_path == "") {
 			printerr ("Path to JavaScript file must be specified\n");
 			return 3;
+		} else if (script_path == "-") {
+			script_path = null;
+			script_source = read_stdin ();
 		}
```
oleavr: How about rephrasing this as:

```vala
		if (script_path == null || script_path == "") {
			printerr ("Path to JavaScript file must be specified\n");
			return 3;
		}

		string? script_source = null;
		if (script_path == "-") {
			script_path = null;
			script_source = read_stdin ();
		}
```

inject/inject.vala
```diff
@@ -261,13 +284,19 @@ namespace Frida.Inject {
...
 			try {
-				var name = Path.get_basename (script_path).split (".", 2)[0];
-
+				string name;
 				string source;
-				try {
-					FileUtils.get_contents (script_path, out source);
-				} catch (FileError e) {
-					throw new Error.INVALID_ARGUMENT (e.message);
+
```
oleavr: Let's drop this blank line so it's clearer that the following logic is responsible for setting up these two variables.

PR #223, opened 2019-01-12
Make injector spawn and attach target if it's not running

iGio90:


oleavr:
Thanks! I think spawning is a nice feature, but I think we should add an explicit switch for it. Especially since the process name and the app ID aren't usually the same on iOS.

oleavr:
Perhaps `-f` for consistency with the other CLI tools in `frida-tools`.

That will only solve the case of wanting to `spawn` though, which means any already running instance gets killed... perhaps we can support the "only spawn if necessary" case by requiring you to specify both `-f` and `-n`.

iGio90:
this should be the easier solution? providing -f for explicit ask for spawn?

inject/inject.vala
```diff
@@ -13,6 +14,7 @@ namespace Frida.Inject {
 		{ "pid", 'p', 0, OptionArg.INT, ref target_pid, null, "PID" },
 		{ "name", 'n', 0, OptionArg.STRING, ref target_name, null, "PID" },
 		{ "script", 's', 0, OptionArg.FILENAME, ref script_path, null, "JAVASCRIPT_FILENAME" },
+		{ "spawn", 'f', 0, OptionArg.NONE, ref spawn, "Spawn target application", null },
```
oleavr: This should be a string, not a boolean, and should be `--file` in long-form, like in frida-tools:

`-f FILE, --file=FILE  spawn FILE`

E.g. `frida-inject -f /bin/cat /etc/hosts`.

Extra arguments should be collected as the arguments to pass.

oleavr:
Timed out. Please reopen later if you're interested in addressing the remaining issues.

PR #230, opened 2019-03-05
Add preliminary support for iOS 12.x

mrmacete:
- avoid calling into SpringBoardServices during app launch (to avoid deadlocking SpringBoard)
- let `need_helpers` be always `TRUE` for now
- include iOS 12.0 in the buggy set because the beta is still buggy
- deal with `SBSCopyFrontmostApplicationDisplayIdentifier` returning garbage of length 1 when no frontmost app is there
- add the `com.apple.private.security.container-required` entitlement set to `false`

For the records: these fixes are jailbreak-agnostic, that means it's just adapting to the newer system - different jailbreaks may need different jailbreak-specific tricks for Frida to properly work there.


src/darwin/frida-helper-backend-glue.m
```diff
@@ -1325,19 +1326,44 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+  if (proc_pidpath (pid, path, sizeof (path)) <= 0)
+    return result;
+
+  dirname = g_path_get_dirname (path);
+  g_strlcpy (path, dirname, sizeof (path));
+  g_free (dirname);
+
+  g_strlcat (path, "/Info.plist", sizeof (path));
+
+  if (stat (path, &info) != 0)
```
oleavr:
```suggestion
  if (!g_file_test (path, G_FILE_TEST_EXISTS))
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1325,19 +1326,44 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
 frida_darwin_helper_backend_is_application_process (guint pid)
 {
   gboolean result = FALSE;
+  gchar path[4 * MAXPATHLEN];
+  gchar * dirname;
+  struct stat info;
```
oleavr:
```suggestion
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1325,19 +1326,44 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+  NSString * identifier = nil;
+
+  if (proc_pidpath (pid, path, sizeof (path)) <= 0)
+    return result;
+
+  dirname = g_path_get_dirname (path);
+  g_strlcpy (path, dirname, sizeof (path));
+  g_free (dirname);
+
+  g_strlcat (path, "/Info.plist", sizeof (path));
```
oleavr: Perhaps it would be simpler to chain `URLByDeletingLastPathComponent` and `URLByAppendingPathComponent:isDirectory:` and let the `plist == nil` error-handling take care of the rest.

src/darwin/system-darwin.m
```diff
@@ -105,6 +105,8 @@
     result->_pid = 0;
     frida_image_data_init (&result->_small_icon, 0, 0, 0, "");
     frida_image_data_init (&result->_large_icon, 0, 0, 0, "");
+    if (identifier != nil)
+      [identifier release];
```
oleavr: We could merge the two `release` statements by putting one after the `if/else`. The `nil`-check isn't necessary though, as it's fine to send any message to `nil`.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1325,19 +1326,37 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
 frida_darwin_helper_backend_is_application_process (guint pid)
 {
   gboolean result = FALSE;
+  gchar path[4 * MAXPATHLEN];
   NSAutoreleasePool * pool;
-  NSString * identifier;
+  NSURL * plist_url;
+  NSDictionary * plist = nil;
+  NSString * identifier = nil;
```
oleavr: These two don't need to be initialized right?

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1325,19 +1326,37 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+
+  if (proc_pidpath (pid, path, sizeof (path)) <= 0)
+    return result;
+
+  if (!g_file_test (path, G_FILE_TEST_EXISTS))
+    return result;
```
oleavr: This check can now be removed, since it's now checking if the program's executable exists.

mrmacete: oops, right!

PR #231, opened 2019-03-21
Get rid of plist check for detecting apps

mrmacete:
Instead parse the executable header and look for dependency on UIKit

depends on https://github.com/frida/frida-gum/pull/329


oleavr:
Tweaked version committed as 35e0c72448462f418fb5557454f92eabceecec28. Thanks! 🤘

PR #234, opened 2019-04-26
Fix CreateProcessInternalW in older windows versions (child_gating)

ineedblood:
Related to https://github.com/frida/frida/issues/854

Frida didn't have a fallback API to call and it was sending a null address to the upcoming calls. This should fix the problem in older versions of Windows where CreateProcessInternalW is in kernel32.dll and not yet in Kernelbase.dll

### DIFF
--- a/lib/payload/spawn-monitor.vala
+++ b/lib/payload/spawn-monitor.vala
@@ -38,11 +38,11 @@ namespace Frida {
 			var interceptor = Gum.Interceptor.obtain ();
 
 #if WINDOWS
-      var cpiw_export = Gum.Module.find_export_by_name ("kernelbase.dll", "CreateProcessInternalW");
-      if (cpiw_export == null)
-        cpiw_export = Gum.Module.find_export_by_name ("kernel32.dll", "CreateProcessInternalW");
-	  assert(cpiw_export != null);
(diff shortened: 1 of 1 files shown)


lib/payload/spawn-monitor.vala
```diff
@@ -38,7 +38,11 @@ namespace Frida {
 			var interceptor = Gum.Interceptor.obtain ();
 
 #if WINDOWS
-			interceptor.attach_listener (Gum.Module.find_export_by_name ("kernelbase.dll", "CreateProcessInternalW"), this);
+			var cpiw_export = Gum.Module.find_export_by_name ("kernelbase.dll", "CreateProcessInternalW");
```
oleavr: Let's name it `cpiw` or `create_process_internal` for consistency with the Darwin code below.

lib/payload/spawn-monitor.vala
```diff
@@ -38,7 +38,11 @@ namespace Frida {
...
+			if (cpiw_export == null)
+				cpiw_export = Gum.Module.find_export_by_name ("kernel32.dll", "CreateProcessInternalW");
+			assert(cpiw_export != null);
```
oleavr: Missing space before `(`.

PR #235, opened 2019-04-28
Add g_set_prgname() to frida_init()

thestr4ng3r:


src/frida-glue.c
```diff
@@ -23,6 +23,7 @@ frida_init (void)
 
   if (g_once_init_enter (&frida_initialized))
   {
+    g_set_prgname("frida");
```
oleavr:
```suggestion
    g_set_prgname ("frida");

```

bet4it:
Do you remember why you need to set it?

thestr4ng3r:
Related conversation: https://t.me/fridadotre/30532

Florian Märkl, [28.04.2019, 17:43:51]:
I'm checking...



Ole André Vadla Ravnås, [28.04.2019, 17:56:34]:
@thestr4ng3r Hm, nice catch! Looks like a bug in one of my GLib patches. Lemme have a closer look

Florian Märkl, [28.04.2019, 17:58:13]:
Yeah it does look wrong. It checks if self->priv->file is NULL and if it is, calls _g_object_unref0 (self->priv->file);

Ole André Vadla Ravnås, [28.04.2019, 17:59:57]:
@thestr4ng3r No, it's crashing here: https://github.com/frida/glib/blob/30920fec4f0a8b9a1578915ba1b9d39d994d6643/gio/win32/gwinhttpvfs.c#L120-L140

@thestr4ng3r Looks like g_get_prgname () is returning NULL

Florian Märkl, [28.04.2019, 18:02:32]:
oh yeah right

Ole André Vadla Ravnås, [28.04.2019, 18:02:51]:
@thestr4ng3r Actually that looks like a bug in GWinHttpVfs (in GLib)... it assumes that g_set_prgname () has been called previously... but, wouldn't hurt to set it anyway I guess. We could add that to frida_init ()

@thestr4ng3r Could you try adding it before this line: https://github.com/frida/frida-core/blob/master/src/frida-glue.c#L26

g_set_prgname ("frida")

@thestr4ng3r And here: https://github.com/frida/frida-gum/blob/master/gum/gum.c#L264

Florian Märkl, [28.04.2019, 18:14:20]:
works!

Only the first didn't but with the second it does.

Ole André Vadla Ravnås, [28.04.2019, 18:15:18]:
Yeah both are needed for different circumstances

Issue #241, opened 2019-05-20
Child gating fails to resume the process after exec() change child process's cpu type


gebing:
I am trying use `.enable_child_gating()` to solve the frida bug of Android will hangs the thread when call `Runtime.exec` with invalid shell commands, and frida will throw `InvalidArgumentError: invalid PID` when i call `self._device.resume(child.pid)`.

Here is my python code for dealing with `.enable_child_gating()` and `resume`

```
self._device.on("child-added", lambda child: self._reactor.schedule(lambda: self._on_child_added(child)))
self._device.on("child-removed", lambda child: self._reactor.schedule(lambda: self._on_child_removed(child)))
self._session.enable_child_gating()
...
def _on_child_added(self, child):
    self._print("Child added: {}".format(child))
    if child.origin != "exec":
        session = self._device.attach(child.pid)
        session.on("detached", lambda reason: self._reactor.schedule(lambda: self._on_child_detached(child.pid, session, reason)))
        session.enable_child_gating()
    self._device.resume(child.pid)

def _on_child_removed(self, child):
    self._print("Child removed: {}".format(child))

def _on_child_detached(self, pid, session, reason):
    self._print("Child detached: pid={}, reason='{}'".format(pid, reason))
```

and the console output is:
```
Child added: Child(pid=2597, parent_pid=2546, origin=fork)
Child removed: Child(pid=2597, parent_pid=2546, origin=fork)
Child detached: pid=2597, reason='process-replaced'
[...]

PR #242, opened 2019-05-23
Fixed for Issue #241

gebing:


oleavr:
Thanks! This is a great find. I'm a bit concerned about caching in this case, as other requests may end up hitting the wrong helper process. What do you think about making the `resume ()` implementation do a try/catch, and checking if `e is Error.INVALID_ARGUMENT` and if that is the case, retry the request on the other helper? (And if not, rethrow.)

gebing:
i think your idea is ok, and i can modify this PR and do some test.

gebing:
I have tested it and it works ok

oleavr:
Looks great, thanks!

PR #247, opened 2019-07-04
Android gadget get executable name and package name

gebing:
The executable name in gadget of android is always 'app_process32' or 'app_process64', and the bundle name is always null. So the feature of script optional configuration in ScriptDirectory mode is not work in android because of the filter will always not match.

I add additional  code for Android to retrieve real executable name and package name, and now android's gadget can work with script optional configuration such as below:

```
{
  "filter": {
     "executables": ["com.target.app:remote"]
  }
}
```

or like:

```
{
  "filter": {
     "bundles": ["com.target.app"]
  }
}
```

### DIFF
--- a/lib/gadget/gadget.vala
+++ b/lib/gadget/gadget.vala
@@ -313,7 +313,7 @@ namespace Frida.Gadget {
 					FileUtils.get_contents ("/proc/self/cmdline", out cmdline);
 					if (cmdline != "zygote" && cmdline != "zygote64") {
 						executable_name = cmdline;
-						cached_bundle_id = cmdline.substring(0, cmdline.last_index_of (":"));
+						cached_bundle_id = cmdline[0:cmdline.last_index_of (":")];
 					}
 				} catch (FileError e) {
 				}
--- a/lib/gadget/gadget.vala
+++ b/lib/gadget/gadget.vala
@@ -304,6 +304,22 @@ namespace Frida.Gadget {
 				range: range
 			);
 		}
+
+#if ANDROID
+		construct {
+			if (executable_name.index_of ("app_process") >= 0) {
+				try {
+					string cmdline;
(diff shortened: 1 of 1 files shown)


lib/gadget/gadget.vala
```diff
@@ -304,6 +304,22 @@ namespace Frida.Gadget {
 				range: range
 			);
 		}
+
+#if ANDROID
+		construct {
+			if (executable_name.index_of ("app_process") >= 0) {
```
oleavr: Would be good to make this check a bit stricter, to avoid false positives:

```suggestion
			if (executable_name.has_prefix ("app_process")) {
```

lib/gadget/gadget.vala
```diff
@@ -304,6 +304,22 @@ namespace Frida.Gadget {
...
+				try {
+					string cmdline;
+					FileUtils.get_contents ("/proc/self/cmdline", out cmdline);
+					if (cmdline != "zygote" && cmdline != "zygote64") {
+						executable_name = cmdline;
+						cached_bundle_id = cmdline.substring(0, cmdline.last_index_of (":"));
```
oleavr: Missing space before opening parenthesis. However, I think this can be simplified by using Vala's string slicing syntax:

```suggestion
						cached_bundle_id = cmdline[0:cmdline.last_index_of (":")];
```

(Untested so would be good to verify that it supports negative indexing.)

xiaobaiyey:
[https://github.com/frida/frida-core/blob/master/lib/gadget/gadget.vala#L316](https://github.com/frida/frida-core/blob/master/lib/gadget/gadget.vala#L316)

`cached_bundle_id = cmdline[0:cmdline.last_index_of (":")];`
cmdline.last_index_of (":") 
for android app will got -1, not all porcess cmdlie file content has ":";
The cached_bundle_id retrieved will be one character short,
so I thought I'd check to see if the return value of last_index_of is equal to -1,

PR #248, opened 2019-07-12
bugfix: app icons on iOS12 are missing

ChiChou:
From iOS12, `SBSCopyIconImagePNGDataForDisplayIdentifier` requires `com.apple.springboard.iconState`, otherwise it would fail


oleavr:
Thanks! Would be good to also add it to the `frida-helper.xcent`, even if it's only used for devkits.

ChiChou:
@oleavr Please don't merge now. The binary format is currently broken

```frida-core/server/frida-server.xcent: invalid length in entitlement blob
make[1]: *** [build/.core-ios-stamp-frida_thin-ios-arm64] Error 1
```

ChiChou:
@oleavr All three xcent files have been updated

ChiChou:
@oleavr All three xcent files have been updated.

Here's a simple script to update the entitlements

```python
import struct
import plistlib


def patch(filename):
    import os
    absname = os.path.join('GitHub/frida', filename)
    with open(absname, 'rb') as fp:
        buf = fp.read()

    magic, length = struct.unpack('>II', buf[0:8])
    print('magic', hex(magic))
    print('len', length)
    assert(len(buf) == length)
    o = plistlib.loads(buf[8:])
    o['com.apple.springboard.iconState'] = True
    # print(o)
    ent = plistlib.dumps(o)
    adjusted = buf[0:4] + struct.pack('>I', len(ent) + 8) + ent
    with open(absname, 'wb') as fp:
        fp.write(adjusted)


todo = [
    './frida-core/tests/frida-tests.xcent',
    './frida-core/server/frida-server.xcent',
    './frida-core/src/darwin/frida-helper.xcent']

for item in todo:
    patch(item)
```

But I think it's better to track the sources in plain text and use a build script to concert them

ChiChou:
The truth is that these files can be in plain text...

oleavr:
Thanks! 🤘

PR #249, opened 2019-07-13
Add --parameters to frida-inject to allow specify parameters using a JSON string similar to frida-gadget

eugenekolo:
My first Vala code, so it might not be vala-thonic

`./frida-inject -p 16920 --script test.js --parameters '{"paramBool": true, "paramArray": [1,2,3], "paramInt": 1, "paramStr": "str", "paramObj": {"a": ["b", "c", {}]}}'`

```
stage early
parameters {"paramBool":true,"paramArray":[1,2,3],"paramInt":1,"paramStr":"str","paramObj":{"a":["b","c",{}]}}
```

### DIFF
--- a/inject/inject.vala
+++ b/inject/inject.vala
@@ -4,6 +4,7 @@ namespace Frida.Inject {
 	private static int target_pid = -1;
 	private static string? target_name;
 	private static string? script_path;
+	private static string? parameters;
 	private static bool eternalize;
 	private static bool enable_jit;
 	private static bool enable_development;
@@ -13,6 +14,7 @@ namespace Frida.Inject {
 		{ "pid", 'p', 0, OptionArg.INT, ref target_pid, null, "PID" },
 		{ "name", 'n', 0, OptionArg.STRING, ref target_name, null, "PID" },
 		{ "script", 's', 0, OptionArg.FILENAME, ref script_path, null, "JAVASCRIPT_FILENAME" },
+		{ "parameters",  's', 0, OptionArg.STRING, ref parameters, "Parameters as JSON, same as Gadget", "PARAMETERS_JSON"},
 		{ "eternalize", 'e', 0, OptionArg.NONE, ref eternalize, "Eternalize script and exit", null },
 		{ "enable-jit", 0, 0, OptionArg.NONE, ref enable_jit, "Enable the JIT runtime", null },
 		{ "development", 'D', 0, OptionArg.NONE, ref enable_development, "Enable development mode", null },
@@ -59,7 +61,7 @@ namespace Frida.Inject {
 			script_source = read_stdin ();
 		}
 
-		application = new Application (target_pid, target_name, script_path, script_source, enable_jit, enable_development);
+		application = new Application (target_pid, target_name, script_path, script_source, parameters, enable_jit, enable_development);
 
 #if !WINDOWS
 		Posix.signal (Posix.Signal.INT, (sig) => {
@@ -117,6 +119,11 @@ namespace Frida.Inject {
 			construct;
 		}
 
+		public string? parameters {
+			get;
+			construct;
+		}
+
 		public bool enable_jit {
 			get;
 			construct;
@@ -134,12 +141,13 @@ namespace Frida.Inject {
 		private MainLoop loop;
 		private bool stopping;
 
-		public Application (int target_pid, string? target_name, string? script_path, string? script_source, bool enable_jit, bool enable_development) {
+		public Application (int target_pid, string? target_name, string? script_path, string? script_source, string? parameters, bool enable_jit, bool enable_development) {
 			Object (
 				target_pid: target_pid,
 				target_name: target_name,
 				script_path: script_path,
 				script_source: script_source,
+				parameters: parameters,
 				enable_jit: enable_jit,
 				enable_development: enable_development
 			);
@@ -175,7 +183,7 @@ namespace Frida.Inject {
 
 				var session = yield device.attach (pid);
 
-				var r = new ScriptRunner (session, script_path, script_source, enable_jit, enable_development);
+				var r = new ScriptRunner (session, script_path, script_source, parameters, enable_jit, enable_development);
 				yield r.start ();
 				script_runner = r;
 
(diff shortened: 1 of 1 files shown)


eugenekolo:
Done. Moved to where the other args are passed.

Tested now with passing blank parameters too. Same as Gadget handling, where it passes a {}.

Testing:
```
kiwi:/data/local/tmp # cat test.js 
'use strict';

rpc.exports = {
    init: function (stage, parameters) {
	console.log('stage', stage);
	console.log('parameters', JSON.stringify(parameters));
    },
};

kiwi:/data/local/tmp # ./frida-inject -p 1256 --script test.js --parameters '{"parmBool": true, "paramArray": [1,2,3], "paramInt": 1, "paramStr": "str", "paramObj": {"a": ["b", "c", {}]}}'                                                    <
stage early
parameters {"paramBool":true,"paramArray":[1,2,3],"paramInt":1,"paramStr":"str","paramObj":{"a":["b","c",{}]}}

kiwi:/data/local/tmp # ./frida-inject -p 1256 --script test.js --parameters ''                                                                                 
stage early
parameters {}

kiwi:/data/local/tmp # ./frida-inject -p 1256 --script test.js --parameters '{}'                                                                               
stage early
parameters {}

kiwi:/data/local/tmp # ./frida-inject -p 1256 --script test.js                                                                                                 
stage early
parameters {}
```

inject/inject.vala
```diff
@@ -59,7 +61,18 @@ namespace Frida.Inject {
 			script_source = read_stdin ();
 		}
 
-		application = new Application (target_pid, target_name, script_path, script_source, enable_jit, enable_development);
+		Json.Node parameters = new Json.Node (Json.NodeType.OBJECT);
```
oleavr:
```suggestion
		var parameters = new Json.Node (Json.NodeType.OBJECT);
```

inject/inject.vala
```diff
@@ -59,7 +61,18 @@ namespace Frida.Inject {
...
+		parameters.set_object (new Json.Object ()); 
+		if (parameters_str != null && parameters_str != "") {
```
oleavr: The empty string case should be an error. We can just ignore that here and let the parsing below fail.

inject/inject.vala
```diff
@@ -59,7 +61,18 @@ namespace Frida.Inject {
...
+			try {
+				var parser = new Json.Parser ();
+				parser.load_from_data (parameters_str, -1);
+				parameters.set_object (parser.get_root().get_object ());
+			} catch (GLib.Error e) {
```
oleavr: Should `printerr ()` here and then return.

inject/inject.vala
```diff
@@ -59,7 +61,18 @@ namespace Frida.Inject {
...
 		}
 
-		application = new Application (target_pid, target_name, script_path, script_source, enable_jit, enable_development);
+		Json.Node parameters = new Json.Node (Json.NodeType.OBJECT);
+		parameters.set_object (new Json.Object ()); 
+		if (parameters_str != null && parameters_str != "") {
+			try {
+				var parser = new Json.Parser ();
+				parser.load_from_data (parameters_str, -1);
+				parameters.set_object (parser.get_root().get_object ());
```
oleavr:
```suggestion
				parameters.set_object (parser.get_root ().get_object ());
```

inject/inject.vala
```diff
@@ -117,6 +130,11 @@ namespace Frida.Inject {
 			construct;
 		}
 
+		public Json.Node? parameters {
```
oleavr: Did you try making this non-nullable?

inject/inject.vala
eugenekolo: The empty string doesn't cause an error if you let it parse. "" is considered valid JSON in glib. 

Added a case for it, although I'm not sure it erroring is any better than just continuing on with an empty object.

inject/inject.vala
```diff
@@ -59,7 +61,25 @@ namespace Frida.Inject {
...
+				printerr ("Parameters argument must be specified as JSON if present\n");
+				return 4;
+			}
+
+			try {
+				var parser = new Json.Parser ();
+				parser.load_from_data (parameters_str, -1);
+				parameters.set_object (parser.get_root ().get_object ());
+			} catch (GLib.Error e) {
+				printerr ("Failed to parse parameters argument as JSON\n");
```
oleavr: Would be good to include the error message:

```suggestion
				printerr ("Failed to parse parameters argument as JSON: %s\n", e.message);
```

inject/inject.vala
```diff
@@ -59,7 +61,25 @@ namespace Frida.Inject {
...
 		}
 
-		application = new Application (target_pid, target_name, script_path, script_source, enable_jit, enable_development);
+		var parameters = new Json.Node (Json.NodeType.OBJECT);
+		parameters.set_object (new Json.Object ()); 
+		if (parameters_str != null) {
+			if (parameters_str == "") {
+				printerr ("Parameters argument must be specified as JSON if present\n");
+				return 4;
+			}
```
oleavr: Why not let it fail below – or is the error message unintuitive?

eugenekolo: It seg faults.. so very unintuitive.

oleavr: 🤣 Sounds like a bug in JSON-GLib. Anyway, let's work around it for now.

inject/inject.vala
```diff
@@ -13,6 +14,7 @@ namespace Frida.Inject {
 		{ "pid", 'p', 0, OptionArg.INT, ref target_pid, null, "PID" },
 		{ "name", 'n', 0, OptionArg.STRING, ref target_name, null, "PID" },
 		{ "script", 's', 0, OptionArg.FILENAME, ref script_path, null, "JAVASCRIPT_FILENAME" },
+		{ "parameters",  's', 0, OptionArg.STRING, ref parameters_str, "Parameters as JSON, same as Gadget", "PARAMETERS_JSON" },
```
oleavr: The short switch conflicts with `--scripts` / `-s`. Perhaps we should use `'P'`?

eugenekolo: Changed to P

oleavr:
Thanks, merged! 🤘

PR #251, opened 2019-08-01
fix some time get cached_bundle_id error

xiaobaiyey:
fix some time get cached_bundle_id error

### DIFF
--- a/lib/gadget/gadget.vala
+++ b/lib/gadget/gadget.vala
@@ -313,7 +313,12 @@ namespace Frida.Gadget {
 					FileUtils.get_contents ("/proc/self/cmdline", out cmdline);
 					if (cmdline != "zygote" && cmdline != "zygote64") {
 						executable_name = cmdline;
-						cached_bundle_id = cmdline[0:cmdline.last_index_of (":")];
+						var index = cmdline.last_index_of (":");
+						if(index==-1){
+						  cached_bundle_id = cmdline;
+						}else{
+						  cached_bundle_id = cmdline[0:index];
+						}
 					}
 				} catch (FileError e) {
 				}


oleavr:
It seems to me that this is only a symptom, and not the cause. Perhaps the correct solution is to get this information from `/proc/self/cmdline`, but instead get it from somewhere where the information is not locale-dependent? I.e. `:` is probably only used for Western locales... so this string is only meant for humans to read, it's not meant to be machine readable.

xiaobaiyey:
when I use Frida gadget's script-directory mode
```
angler:/ $ ps -A | grep com.coolapk.market
u0_a93       17574   610 1942552 176380 SyS_epoll_wait      0 S com.coolapk.market
u0_a93       17672   610 1699444  43636 SyS_epoll_wait      0 S com.coolapk.market:pushservice
```
in this case,**com.coolapk.market** has two processes，one is **com.coolapk.market** ,the other is **com.coolapk.market:pushservice**

In the configuration file of the javascript.
**executables** should fill in the full process name
**bundles** should fill in the package name like iOS **BundleID**


* for **com.coolapk.market:pushservice**
we got **/proc/17672/cmdline** content
```
angler:/ $ cat  /proc/17672/cmdline
com.coolapk.market:pushservice
```
we got executable_name  is **com.coolapk.market:pushservice**
and cached_bundle_id is **com.coolapk.market**
The result is correct.

* for **com.coolapk.market**
we got **/proc/17574/cmdline** content
```
angler / # cat /proc/17574/cmdline 
com.coolapk.market
```
we got executable_name  is **com.coolapk.market**
and cached_bundle_id is **com.coolapk.marke**(You should notice  there is a missing letter t here)

sometime we set bundles with "com.coolapk.market",the configuration file will not work,because the cached_bundle_id we got is wrong，one character is missing。

If we fill "com.coolapk.market" in both executables and bundles, it will not affect normal use.
That's my understanding.
haha😂😂

oleavr:
Ahh, sorry, I misunderstood. So your fix was good then :-) Sorry for the confusion!

I'm releasing 12.6.12 now so I fixed this quickly in 244a47ec32f85d85e236dba1d5b66ddd7f574565.

Let me know how this new release works for you. Thanks!

PR #264, opened 2019-10-07
Split the state by thread id in the ReportCrash agent

mrmacete:
Because in some cases more than one crash can be processed in parallel in different threads.

### DIFF
--- a/src/darwin/agent/reportcrash.js
+++ b/src/darwin/agent/reportcrash.js
@@ -51,10 +51,10 @@ var AppleErrorReport = ObjC.classes.AppleErrorReport;
 var CrashReport = ObjC.classes.CrashReport;
 var NSMutableDictionary = ObjC.classes.NSMutableDictionary;
 
-const state = {};
+var threadStates = {};
 
 function reset(threadId) {
-  const localState = {
+  var state = {
     crashedPid: -1,
     is64Bit: null,
     forcedByUs: false,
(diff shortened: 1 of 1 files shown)


src/darwin/agent/reportcrash.js
```diff
@@ -96,45 +97,70 @@ Interceptor.attach(Module.getExportByName(CORESYMBOLICATION_PATH, 'task_is_64bit
...
-      unlink(Memory.allocUtf8String(logPath));
-      reset();
+    const localState = state[this.threadId];
+    if (localState === undefined) {
+      console.error('- saveToDir: missing state for thread', this.threadId);
+      return;
+    }
+    if (localState.forcedByUs) {
+      unlink(Memory.allocUtf8String(localState.logPath));
+      delete state[this.threadId];
```
mrmacete: i wonder if it's better to move this outside the `if`

src/darwin/agent/reportcrash.js
```diff
@@ -145,30 +171,49 @@ Interceptor.attach(Module.getExportByName(LIBSYSTEM_KERNEL_PATH, 'open_dprotecte
...
-    if (this.isCrashLog)
-      logFd = retval.toInt32();
+    const localState = state[this.threadId];
+    if (localState === undefined) {
+      console.error('open_dprotected_np: missing state for thread', this.threadId);
+      return;
+    }
+    if (this.isCrashLog) {
+      localState.logFd = retval.toInt32();
+    }
```
oleavr: Brackets should be omitted for consistency.

src/darwin/agent/reportcrash.js
```diff
@@ -51,35 +51,36 @@ var AppleErrorReport = ObjC.classes.AppleErrorReport;
...
-function reset() {
-  crashedPid = -1;
-  is64Bit = null;
-  forcedByUs = false;
-  logPath = null;
-  logFd = null;
-  logChunks = [];
-  mappedAgents = [];
-}
+const state = {};
```
oleavr: - Even though newer Duktape (kinda) supports `const`, we should probably stick to `var` for consistency with our existing internal agents written in ES5 syntax.
- How about naming it `threadStates`, and use `state` for the local cases? (So the variable name length relationships are proportional to their visibility.)

src/darwin/agent/reportcrash.js
```diff
@@ -96,45 +97,70 @@ Interceptor.attach(Module.getExportByName(CORESYMBOLICATION_PATH, 'task_is_64bit
     this.pid = pidForTask(args[0].toUInt32());
   },
   onLeave: function (retval) {
-    if (this.pid === crashedPid)
-      is64Bit = !!retval.toUInt32();
+    const localState = state[this.threadId];
+    if (localState === undefined) {
+      console.error('task_is_64bit: missing state for thread', this.threadId);
```
oleavr: Hmm, how about making a `getState(operation)`, where `operation` is e.g. `task_is_64bit`, that throws an exception formatted exactly like this when `undefined`. Unhandled JS exceptions are logged to `stderr`, so the result will be roughly the same, but will simplify the code a little bit.

src/darwin/agent/reportcrash.js
oleavr: Yep I think that's a good idea.

PR #268, opened 2019-10-09
Keep eternalized entries alive

mrmacete:
So that they can be reused for future connections to the same process, avoiding the need to have more than one agent, at least during the lifetime of frida-server.


oleavr:
Awesome! ❤️

PR #270, opened 2019-10-10
Allow launchd agent to kill pending suspended processes on dispose

mrmacete:
In this way if frida-server dies it doesn’t leave back unattended pending spawns in a suspended state.

### DIFF
--- a/src/darwin/agent/launchd.js
+++ b/src/darwin/agent/launchd.js
@@ -1,6 +1,7 @@
 var pointerSize = Process.pointerSize;
 
 var POSIX_SPAWN_START_SUSPENDED = 0x0080;
+var SIGKILL = 9;
 
 var upcoming = {};
 var gating = false;
@@ -9,6 +10,8 @@ var jbdPidsToIgnore = null;
 
 var substrateInvocations = {};
 var substratePidsPending = {};
+
+var suspendedPids = {};
 
 rpc.exports = {
   prepareForLaunch: function (identifier) {
@@ -24,6 +27,16 @@ rpc.exports = {
   disableSpawnGating: function () {
     gating = false;
   },
+  forgetPid: function (pid) {
+    delete suspendedPids[pid];
+  },
+  dispose: function () {
+    var kill = new NativeFunction(Module.getExportByName(null, 'kill'));
+    Object.keys(suspendedPids)
+      .forEach(function (pid) {
+        kill (pid, SIGKILL);
+      });
+  }
 };
 
 applyJailbreakQuirks();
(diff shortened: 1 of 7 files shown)


src/darwin/agent/launchd.js
```diff
@@ -24,6 +27,16 @@ rpc.exports = {
...
     gating = false;
   },
+  forgetPid: function (pid) {
+    delete suspendedPids[pid];
+  },
+  dispose: function () {
+    var kill = new NativeFunction(Module.getExportByName(null, 'kill'));
+    Object.keys(suspendedPids)
+      .forEach(function (pid) {
+        kill (pid, SIGKILL);
```
oleavr:
```suggestion
        kill(pid, SIGKILL);
```

oleavr: How does this work, isn't `pid` a string as it's coming from `Object.keys()`?

src/darwin/agent/launchd.js
```diff
@@ -24,6 +27,16 @@ rpc.exports = {
   disableSpawnGating: function () {
     gating = false;
   },
+  forgetPid: function (pid) {
+    delete suspendedPids[pid];
+  },
+  dispose: function () {
+    var kill = new NativeFunction(Module.getExportByName(null, 'kill'));
```
oleavr: This call seems to be missing two arguments (`returnType` and `argTypes`).

src/darwin/agent/launchd.js
```diff
@@ -24,6 +27,16 @@ rpc.exports = {
   disableSpawnGating: function () {
     gating = false;
   },
+  forgetPid: function (pid) {
+    delete suspendedPids[pid];
+  },
+  dispose: function () {
```
oleavr: Would be good to move this first for consistency (the convention followed currently is to always put `dispose` first, or, right after the constructor).

src/darwin/darwin-host-session.vala
```diff
@@ -572,6 +576,14 @@ namespace Frida {
 			handle_spawn.begin (info);
 		}
 
+		private void on_resumed (uint pid) {
+			launchd_agent.forget_pid.begin (pid, io_cancellable);
+		}
+
+		private void on_killed (uint pid) {
+			launchd_agent.forget_pid.begin (pid, io_cancellable);
+		}
```
oleavr: Perhaps these would read slightly better as `on_process_resumed` and `on_process_killed`?

src/darwin/agent/launchd.js
```diff
@@ -24,6 +27,16 @@ rpc.exports = {
   disableSpawnGating: function () {
     gating = false;
   },
+  forgetPid: function (pid) {
```
oleavr: Perhaps `acknowledgeProcess` would be more descriptive?

oleavr: Or perhaps `notifyProcessHandled`?

src/darwin/agent/launchd.js
```diff
@@ -10,7 +11,16 @@ var jbdPidsToIgnore = null;
 var substrateInvocations = {};
 var substratePidsPending = {};
 
+var suspendedPids = {};
```
oleavr: Sorry I missed this: let's move it right after `gating`.

PR #271, opened 2019-10-10
improve spawn in Android apps

muhzii:
allow using the activity param in case the app doesn't have a launcher activity

### DIFF
--- a/src/linux/agent/system-server.js
+++ b/src/linux/agent/system-server.js
@@ -1,4 +1,4 @@
-var ApplicationInfo, ComponentName, Intent, RunningAppProcessInfo, RunningTaskInfo, GET_META_DATA;
+var ApplicationInfo, ComponentName, Intent, RunningAppProcessInfo, RunningTaskInfo, GET_META_DATA, GET_ACTIVITIES;
 var context, packageManager, activityManager;
 
 var pendingLaunches = {};
@@ -15,6 +15,7 @@ function init() {
   RunningTaskInfo = Java.use('android.app.ActivityManager$RunningTaskInfo');
   var ACTIVITY_SERVICE = Context.ACTIVITY_SERVICE.value;
   GET_META_DATA = PackageManager.GET_META_DATA.value;
+  GET_ACTIVITIES = PackageManager.GET_ACTIVITIES.value;
 
   context = ActivityThread.currentApplication();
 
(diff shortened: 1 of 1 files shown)


src/linux/agent/system-server.js
```diff
@@ -85,12 +86,35 @@ rpc.exports = {
   },
   startActivity: function (pkg, activity) {
     return performOnJavaVM(function () {
-      var intent = packageManager.getLaunchIntentForPackage(pkg);
-      if (intent === null)
+      var appInstalled = false;
+      var apps = packageManager.getInstalledApplications(GET_META_DATA);
+      for (var i = 0; i < apps.size(); i++) {
```
oleavr: Let's only read the `size` once, by moving it above the loop:

```suggestion
      var numApps = apps.size();
      for (var i = 0; i !== numApps; i++) {
```

(And also use the same loop idiom as elsewhere for consistency.)

src/linux/agent/system-server.js
```diff
@@ -85,12 +86,35 @@ rpc.exports = {
...
+          break;
+        }
+      }
+      if (!appInstalled)
         throw new Error("Unable to find application with identifier '" + pkg + "'");
 
-      if (activity !== null)
+      var intent = packageManager.getLaunchIntentForPackage(pkg);
+      if (intent === null && activity === null)
+        throw new Error("Unable to find a front-door activity");
```
oleavr: Strings should be this style:

```suggestion
        throw new Error('Unable to find a front-door activity');
```

(Though we've made a few exceptions in this file for cases where we want to use single-quotes in the string literal.)

src/linux/agent/system-server.js
```diff
@@ -85,12 +86,35 @@ rpc.exports = {
...
+
+      if (intent === null) {
+        intent = Intent.$new();
+        intent.setAction("android.intent.action.MAIN");
```
oleavr:
```suggestion
        intent.setAction('android.intent.action.MAIN');
```

src/linux/agent/system-server.js
```diff
@@ -85,12 +86,35 @@ rpc.exports = {
...
+      }
+
+      if (activity !== null) {
+        var pkgInfo = packageManager.getPackageInfo(pkg, GET_ACTIVITIES);
+        var activities = pkgInfo.activities.value.map((activityInfo) => activityInfo.name.value);
```
oleavr:
```suggestion
        var activities = pkgInfo.activities.value.map((activityInfo) => activityInfo.name.value);
```

Should stick to ES5 in the internal agents so they can also be used with Duktape. (Even if we currently use V8 for this particular agent.)

src/linux/agent/system-server.js
```diff
@@ -85,12 +86,35 @@ rpc.exports = {
...
+        if (activities.indexOf(activity) < 0)
```
oleavr:
```suggestion
        if (activities.indexOf(activity) === -1)
```

(For consistency.)

src/linux/agent/system-server.js
```diff
@@ -100,17 +101,20 @@ rpc.exports = {
...
       }
 
       if (activity !== null) {
+        var activities = [];
         var pkgInfo = packageManager.getPackageInfo(pkg, GET_ACTIVITIES);
-        var activities = pkgInfo.activities.value.map((activityInfo) => activityInfo.name.value);
-        if (activities.indexOf(activity) < 0)
+        pkgInfo.activities.value.forEach(function (activityInfo) {
+          activities.push(activityInfo.name.value);
+        });
```
oleavr: Sorry, I wasn't clear – `map()` is available, only the arrow function part needed changing.

muhzii: of course

PR #274, opened 2019-11-07
Support both cleartext and encrypted lockdown channels

mrmacete:
By default the channel is encrypted, unless the service name includes the “?tls=handshake-only” suffix. Internally, the “com.apple.debugserver” and “com.apple.instruments.remoteserver” channels are kept cleartext.

### DIFF
--- a/src/fruity/fruity-host-session.vala
+++ b/src/fruity/fruity-host-session.vala
@@ -296,14 +296,23 @@ namespace Frida {
 			if (address.has_prefix ("lockdown:")) {
 				string service_name = address.substring (9);
 
-				var client = yield get_lockdown_client (cancellable);
+				if (service_name.length != 0) {
+					var client = yield get_lockdown_client (cancellable);
 
-				try {
-					return yield client.start_service (service_name, cancellable);
-				} catch (GLib.Error e) {
-					if (e is Fruity.LockdownError.INVALID_SERVICE)
+					try {
+						return yield client.start_service (service_name, cancellable);
+					} catch (GLib.Error e) {
+						if (e is Fruity.LockdownError.INVALID_SERVICE)
+							throw new Error.NOT_SUPPORTED ("%s", e.message);
+						throw new Error.TRANSPORT ("%s", e.message);
+					}
+				} else {
+					try {
+						var client = yield Fruity.LockdownClient.open (device_details, cancellable);
+						return client.stream;
+					} catch (GLib.Error e) {
 						throw new Error.NOT_SUPPORTED ("%s", e.message);
-					throw new Error.TRANSPORT ("%s", e.message);
+					}
 				}
 			}
 
--- a/src/fruity/fruity-host-session.vala
+++ b/src/fruity/fruity-host-session.vala
@@ -691,15 +691,13 @@ namespace Frida {
 						argv += provided_argv[i];
 				}
 
-				var lldb_stream = yield lockdown.start_service (DEBUGSERVER_SERVICE_NAME, cancellable);
-				var lldb = yield LLDB.Client.open (lldb_stream, cancellable);
+				var lldb = yield start_lldb_service (lockdown, cancellable);
 				var process = yield lldb.launch (argv, launch_options, cancellable);
 				if (process.observed_state == ALREADY_RUNNING) {
 					yield lldb.kill (cancellable);
 					yield lldb.close (cancellable);
 
-					lldb_stream = yield lockdown.start_service (DEBUGSERVER_SERVICE_NAME, cancellable);
-					lldb = yield LLDB.Client.open (lldb_stream, cancellable);
+					lldb = yield start_lldb_service (lockdown, cancellable);
 					process = yield lldb.launch (argv, launch_options, cancellable);
 				}
 
@@ -708,12 +706,6 @@ namespace Frida {
 
 				return process.pid;
 			} catch (Fruity.InstallationProxyError e) {
-				throw new Error.NOT_SUPPORTED ("%s", e.message);
-			} catch (Fruity.LockdownError e) {
-				if (e is Fruity.LockdownError.INVALID_SERVICE) {
-					throw new Error.NOT_SUPPORTED ("This feature requires an iOS Developer Disk Image to be mounted; " +
-						"run Xcode briefly or use ideviceimagemounter to mount one manually");
-				}
 				throw new Error.NOT_SUPPORTED ("%s", e.message);
 			} catch (LLDB.Error e) {
 				throw new Error.NOT_SUPPORTED ("%s", e.message);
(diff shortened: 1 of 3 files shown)


src/fruity/fruity-host-session.vala
```diff
@@ -393,7 +393,7 @@ namespace Frida {
 		private const uint16 DEFAULT_SERVER_PORT = 27042;
 		private const double MIN_SERVER_CHECK_INTERVAL = 5.0;
 		private const string GADGET_APP_ID = "re.frida.Gadget";
-		private const string DEBUGSERVER_SERVICE_NAME = "com.apple.debugserver";
+		private const string DEBUGSERVER_SERVICE_NAME = "com.apple.debugserver?tls=handshake-only";
```
oleavr: Would be good to move it to where it's used, so this remains just the name of the service.

src/fruity/lockdown.vala
```diff
@@ -75,7 +75,19 @@ namespace Frida.Fruity {
 			closed ();
 		}
 
-		public async IOStream start_service (string name, Cancellable? cancellable = null) throws LockdownError, IOError {
+		public async IOStream start_service (string name_with_options, Cancellable? cancellable = null) throws LockdownError, IOError {
+			bool tls_handshake_only = false;
+			string name = name_with_options;
+
+			var options_start = name_with_options.index_of_char('?');
```
oleavr:
```suggestion
			var options_start = name_with_options.index_of_char ('?');
```

src/fruity/lockdown.vala
```diff
@@ -75,7 +75,19 @@ namespace Frida.Fruity {
...
 		}
 
-		public async IOStream start_service (string name, Cancellable? cancellable = null) throws LockdownError, IOError {
+		public async IOStream start_service (string name_with_options, Cancellable? cancellable = null) throws LockdownError, IOError {
+			bool tls_handshake_only = false;
+			string name = name_with_options;
+
+			var options_start = name_with_options.index_of_char('?');
+			if (options_start != -1) {
+				if (options_start < name_with_options.char_count() - 1) {
```
oleavr: `char_count()` returns the number of characters, but `index_of_char ()` is the index in bytes.

oleavr:
```suggestion
				if (options_start < name_with_options.char_count () - 1) {
```

src/fruity/lockdown.vala
```diff
@@ -75,7 +75,19 @@ namespace Frida.Fruity {
...
+					string options_string = name_with_options.substring(options_start + 1);
```
oleavr:
```suggestion
					string options_string = name_with_options.substring (options_start + 1);
```

src/fruity/lockdown.vala
```diff
@@ -75,7 +75,19 @@ namespace Frida.Fruity {
...
+					tls_handshake_only = options_string == "tls=handshake-only";
+				}
+				name = name_with_options.substring(0, options_start);
```
oleavr:
```suggestion
				name = name_with_options.substring (0, options_start);
```

src/fruity/lockdown.vala
```diff
@@ -75,7 +75,19 @@ namespace Frida.Fruity {
...
+			}
```
oleavr: Hmm, maybe it's simpler to do something like:

```Vala
var tokens = name_with_options.split ("?", 2);
unowned string name = tokens[0];
bool tls_handshake_only = false;
if (tokens.length > 1) {
	unowned string options = tokens[1];
	tls_handshake_only = options == "tls=handshake-only"
}
```

oleavr: (Untested, but hopefully not too far away from something that compiles.)

src/fruity/fruity-host-session.vala
mrmacete: cool! yeah i always underestimate how smart the vala compiler actually is!

src/fruity/fruity-host-session.vala
```diff
@@ -790,7 +790,7 @@ namespace Frida {
 
 			try {
 				var lockdown = yield lockdown_provider.get_lockdown_client (cancellable);
-				var lldb_stream = yield lockdown.start_service (DEBUGSERVER_SERVICE_NAME, cancellable);
+				var lldb_stream = yield lockdown.start_service (DEBUGSERVER_SERVICE_NAME + "?tls=handshake-only", cancellable);
```
oleavr: Hmm, perhaps we should move these operations into a function named something like `start_lldb_service ()`:

```Vala
var lldb_stream = yield lockdown.start_service (DEBUGSERVER_SERVICE_NAME + "?tls=handshake-only", cancellable);
var lldb = yield LLDB.Client.open (lldb_stream, cancellable);
```

(I've already violated DRY pretty bad here 😅)

src/fruity/lockdown.vala
```diff
@@ -77,15 +77,11 @@ namespace Frida.Fruity {
 
 		public async IOStream start_service (string name_with_options, Cancellable? cancellable = null) throws LockdownError, IOError {
 			bool tls_handshake_only = false;
```
oleavr: How about moving this right after `name`, as it's lower level in a sense?

mrmacete: ah yeah, makes sense. i've put it at the beginning because it's like the "outcome" of the whole block of code.

src/fruity/fruity-host-session.vala
```diff
@@ -812,6 +809,12 @@ namespace Frida {
 
 			return yield attach_via_gadget (pid, gadget_details, cancellable);
 		}
+		
+		private async LLDB.Client start_lldb_service (Fruity.LockdownClient lockdown, Cancellable? cancellable)
+				throws Error, IOError, Fruity.LockdownError, LLDB.Error {
```
oleavr:
```suggestion
				throws Fruity.LockdownError, LLDB.Error, IOError {
```

- Assuming `Error` is not thrown here.
- Moving IOError last as it's only for when `cancellable` causes `IOError.CANCELLED`.

src/fruity/fruity-host-session.vala
```diff
@@ -299,7 +299,10 @@ namespace Frida {
 				var client = yield get_lockdown_client (cancellable);
 
 				try {
-					return yield client.start_service (service_name, cancellable);
+					if (service_name.length != 0)
+						return yield client.start_service (service_name, cancellable);
+					else
+						return client.stream;
```
oleavr: Here we have to create a brand new client and return its `stream`, because otherwise we risk that other API calls use the same shared client to start services while the application is using it for other things.

mrmacete: ohh right, ok this needs also lifecycle management i guess, then.

src/fruity/lockdown.vala
```diff
@@ -8,6 +8,9 @@ namespace Frida.Fruity {
 		}
 
 		private PlistServiceClient service;
+		public IOStream stream {
+			get { return service.stream; }
+		}
```
oleavr: Would be good to move this right after `device_details` (with one blank line between), as public signals/properties should be before private members.

src/fruity/fruity-host-session.vala
oleavr: Lifecycle management isn't needed. Here we get a reference on the underlying `IOStream`, and right after the `LockdownClient` ref count will drop to zero, which means also its `PlistServiceClient` ref count will drop to zero, and the `IOStream` ref left will be held here. Because `LockdownClient` and `PlistServiceClient` only interact with the stream when asked to, and don't do anything to it during dispose/finalize, it is perfectly safe to hold on to the `IOStream` but let those other two objects die.

src/fruity/fruity-host-session.vala
```diff
@@ -296,14 +296,24 @@ namespace Frida {
...
-					if (e is Fruity.LockdownError.INVALID_SERVICE)
+					try {
+						return yield client.start_service (service_name, cancellable);
+					} catch (GLib.Error e) {
+						if (e is Fruity.LockdownError.INVALID_SERVICE)
+							throw new Error.NOT_SUPPORTED ("%s", e.message);
+						throw new Error.TRANSPORT ("%s", e.message);
+					}
+				}
+				else {
```
oleavr: `} else {`

mrmacete: oops! 😅

PR #275, opened 2019-11-14
Armbe8

jonwilson030981:
I have added support for ARMBE8. This architecture has little endian code, but big endian data. ARMBE32 has big endian code and data, but is considered legacy, accordingly I have not implemented it.

Also, TinyCC doesn't appear to support BE8, so that isn't implemented for now.


lib/agent/meson.build
```diff
@@ -28,13 +28,19 @@ if host_os_family != 'darwin'
   operations += ['--move', 'destructor', 'frida_deinit_memory', 'first']
 endif
 
+if host_machine.endian() == 'little'
+  endian_option = ['--endian', 'little']
+else
+  endian_option = ['--endian', 'big']
+endif
```
oleavr: Can't we detect this by looking at the `ELF` like we're already doing in modulate.py? (We're already sniffing it to figure out the architecture type.)

tools/modulate.py
```diff
@@ -32,6 +32,9 @@ def main():
     parser.add_argument("--install-name-tool", metavar="/path/to/install_name_tool", type=str, default=None)
     parser.add_argument("--strip", metavar="/path/to/strip", type=str, default=None)
 
+    the_endians = ('big', 'little')
+    parser.add_argument('--endian',  metavar=("|".join(the_endians)), type=str, default='little', choices=the_endians)
```
oleavr: Strings should be double-quoted, except for enum values, which should be single-quoted. (So `'big'` and `'little'` are correct.)

lib/agent/meson.build
jonwilson030981: I did have a look at this, but it seems that objdump doesn't explicitly list the endianness. It instead list the architecture, but isn't very consistent in the naming. It seems different architectures has very different output. Also I wasn't sure how stable those strings would be between different versions of objdump. Happy to change to a whitelist type approach if you think it will work though?

Otherwise, I though of looking at using pyelftools to parse the elf properly, but this would add a dependency.

test-armbe8:     file format elf32-bigarm
architecture: armv6, flags 0x00000112:
EXEC_P, HAS_SYMS, D_PAGED
start address 0x0001070c

/test-arm:     file format elf32-littlearm
architecture: armv5t, flags 0x00000112:
EXEC_P, HAS_SYMS, D_PAGED
start address 0x00010698

/bin/ls:     file format elf64-x86-64
architecture: i386:x86-64, flags 0x00000150:
HAS_SYMS, DYNAMIC, D_PAGED
start address 0x0000000000005850

tools/modulate.py
jonwilson030981: Is this the change you meant? Did I make the change correctly? Not an expert with github!

lib/agent/meson.build
oleavr: Ah ok, let's do it this way for now.

PR #279, opened 2019-11-23
Fixed #269: Parent file descriptors not available in child process 

### LINKED ISSUE #269: Parent file descriptors not available in child process (Mac OS)
It seems that the FD opened by a parent process are being closed by frida somehow and therefore not available on the child/target process instrumented by frida itself.

I attached a [testcase.zip](https://github.com/frida/frida-core/files/3709116/testcase.zip) to demonstrate the issue. The `createfd.c` just create a file descriptor (1234) by duplicating its stdout and the `checkfd.c` just checks that such FD is available via `fcntl`.

Running the sample directly:
```
$ ./createfd ./checkfd
Running ./checkfd...
--- [+] FD 1234 is READY!
```
If we run those with frida in the middle the `checkfd` process won't have access at the FD.
```
$ ./createfd `which frida` --no-pause -q ./checkfd
Running /usr/local/bin/frida...
Spawned `./checkfd`. Resuming main thread!                              
--- FD 1234 is NOT ready!
```
I checked the source code and noticed that for darwin when spawning the process the following flag is set: `POSIX_SPAWN_CLOEXEC_DEFAULT`. According to the documentation: 

> Apple Extension: If this bit is set, then only file descriptors explicitly described by the file_actions argument are available in the spawned process; all of the other file descriptors are automatically closed in the spawned process.

I immediately thought that this flag was the cause of the issue (why is there btw?). 
I patched frida (by removing such flag) and rebuild the framework but I still get the same issue. I then tried to create a "mocked" version of frida where I only start the process the same way it does with `posix_spawn` and with the very same flags and setup (sigset, fds) and in such case it works:

```
$ ./createfd ./fridamock ./checkfd
Running ./fridamock...
- [+] FD 1234 ready in fridamock!
- posix_spawn --
- Child pid: 48996
- Child resumed!
--- [+] FD 1234 is READY!
- Child exited with status 0
```

Can someone please shine some light?

Additional: Frida version 12.7.6
Mac Mojave: 10.14.6

wizche:
Modified the flags for when a new process for target and helper is started. 
This way FDs opened by parent are not lost. This only affect MacOS (darwin).
Check [issue](https://github.com/frida/frida-core/issues/269) for additional details.

### DIFF
--- a/src/darwin/frida-helper-backend-glue.m
+++ b/src/darwin/frida-helper-backend-glue.m
@@ -633,7 +633,7 @@ _frida_darwin_helper_backend_spawn (FridaDarwinHelperBackend * self, const gchar
   sigemptyset (&signal_mask_set);
   posix_spawnattr_setsigmask (&attributes, &signal_mask_set);
 
-  flags = POSIX_SPAWN_SETPGROUP | POSIX_SPAWN_SETSIGMASK | POSIX_SPAWN_CLOEXEC_DEFAULT | POSIX_SPAWN_START_SUSPENDED;
+  flags = POSIX_SPAWN_SETPGROUP | POSIX_SPAWN_SETSIGMASK | POSIX_SPAWN_START_SUSPENDED;
 
   stdio = frida_host_spawn_options_get_stdio (options);
   switch (stdio)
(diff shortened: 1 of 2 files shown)


oleavr:
Good catch, thanks for digging into this!

PR #280, opened 2019-12-12
frida-gadget: support for init-style scripting with "listen"

ChaosData:
```
frida-gadget: listen interaction config support for staving off resume() until
              after script has run and eternalizing

This patch adds two knobs to the frida-gadget config for the "listen" type:

- One enables configuring when the call to resume() is made that will resume
  the program to enable pushing it until client disconnect.
- The other configures each script to be "eternalized" so that they (and the
  instrumentations made by them) stay loaded indefinitely, even after the
  client disconnects.

These are necessary for setting up rpc.exports "init"-style logic for "listen",
which is useful when the program is short-lived (e.g. `ls`). In such case, two
issues currently arise (that this patch seeks to adress):

- The process exits during (or before!) the script's load/execution
  (`"on_load": "wait"` only ensures that resume() is not called before a
  _connection_ is made.)
- Scripts themselves appear to lack a proper mechanism to call resume() on
  their own, requiring it to be triggered on client diconnect. However, this is
  a catch-22 as client disconnects trigger script unloading.

With the patch applied, these cease to be breaking issues. It's not ideal to
have to disconnect to start, but I don't see another way at the moment without
much more invasive changes. This means no `console.log()` after "init", but
`new File("...", "ab")` works just fine.

Note: Technically, custom clients can call `ClientSession::eternalize_script`,
but the standard frida-tools `frida` CLI tool does not have support for it and
the logic for "when" to enable it can be a bit complicated depending on the use
case. To get around this, I've placed the eternalize call in
`ClientSession::destroy_script` (and disabled destroying the script when the
setting is enabled). This has a major benefit in that the interactive console
interface works when the setting is enabled and all instrumentations applied
through it will persist at runtime.
```

I've been doing some work with short-lived binaries through the gadget and wanted a way to get interactive with it. This patch more or less does that.

I've tried to copy the overall code style and avoid any significant changes around stashing state (i.e. in Server as opposed to Client/ClientSession via `server.config`) related to these options (`"on_attach": ("wait"|"attach")` and `"eternalize": ("yes"|"no")`). The names and the values are not necessarily the best ("on_attach" could be something more like `"resume_when": ("load"|"attach"|"disconnect")` which would also take over for `"on_load"`), but I figured I'd try to avoid coining my own terminology and rocking the boat too much. Feel free to rename whatever.

These changes enable the following:

```bash
$ # terminal A
$ cat build/frida-linux-x86_64/lib/frida-gadget.config
{
  "interaction": {
    "type": "listen",
    "address": "127.0.0.1",
    "port": 27042,
    "on_load": "wait",
    "on_attach": "wait",
    "eternalize": "yes"
  }
}
$ LD_PRELOAD=./build/frida-linux-x86_64/lib/frida-gadget.so ls //
[Frida INFO] Listening on 127.0.0.1 TCP port 27042
...
```

```bash
$ # terminal B (somewhere)
$ frida -H 127.0.0.1 -n Gadget
     ____
    / _  |   Frida 12.7.26 - A world-class dynamic instrumentation toolkit
   | (_| |
    > _  |   Commands:
   /_/ |_|       help      -> Displays the help system
   . . . .       object?   -> Display information about 'object'
   . . . .       exit/quit -> Exit
   . . . .
   . . . .   More info at https://www.frida.re/docs/home/
                                                                                
[Remote::Gadget]-> Interceptor.attach(Module.getExportByName('libc-2.27.so', 'opendir'), { 
  onEnter: function (args) { 
    var path = args[0].readUtf8String(); 
    var f = new File("/tmp/frida-opendir.log", "ab"); 
    f.write(path + "\n"); 
    f.flush(); 
    f.close(); 
  }, 
  onLeave: function (retval) { 
  } 
});
{}
[Remote::Gadget]->

Thank you for using Frida!
```

```bash
bin   cdrom  etc   lib    lib64   lost+found  mnt  proc  run   ...
$ # terminal A
$ cat /tmp/frida-opendir.log 
//
```

N.B. It might be better to add an eternalize flag to the `frida` CLI tool, but I figured this was useful enough in gadget since it might be a bit complicated wrt where to add the call (see above commit msg for more details).

### DIFF
--- a/lib/gadget/gadget.vala
+++ b/lib/gadget/gadget.vala
@@ -271,6 +271,18 @@ namespace Frida.Gadget {
 			default = LoadBehavior.WAIT;
 		}
 
+		public AttachBehavior on_attach {
+			get;
+			set;
+			default = AttachBehavior.RESUME;
+		}
+
+		public EternalizeBehavior eternalize {
+			get;
+			set;
+			default = EternalizeBehavior.NO;
+		}
+
 		public enum PortConflictBehavior {
 			FAIL,
 			PICK_NEXT
@@ -279,6 +291,16 @@ namespace Frida.Gadget {
 		public enum LoadBehavior {
 			RESUME,
 			WAIT
+		}
+
+		public enum AttachBehavior {
+			RESUME,
+			WAIT
+		}
+
+		public enum EternalizeBehavior {
+			NO,
+			YES
 		}
 	}
 
@@ -1385,6 +1407,11 @@ namespace Frida.Gadget {
 				server = s;
 				connection = c;
 
+				var listen_interaction = server.config.interaction as ListenInteraction;
+				if (listen_interaction != null) {
+					resume_on_attach = listen_interaction.on_attach == ListenInteraction.AttachBehavior.RESUME;
+				}
+
 				try {
 					host_registration_id = connection.register_object (Frida.ObjectPath.HOST_SESSION,
 						this as HostSession);
@@ -1514,6 +1541,10 @@ namespace Frida.Gadget {
 				session.closed.disconnect (on_session_closed);
 				sessions.remove (session);
 
+				if (!resume_on_attach) {
+					Frida.Gadget.resume ();
+				}
+
 				agent_session_destroyed (session.id, APPLICATION_REQUESTED);
 			}
 
@@ -1610,7 +1641,6 @@ namespace Frida.Gadget {
 			public async AgentScriptId create_script_with_options (string source, AgentScriptOptions options,
 					Cancellable? cancellable) throws Error, IOError {
 				check_open ();
-
 				var instance = yield script_engine.create_script (source, null, ScriptOptions._deserialize (options.data));
 				return instance.script_id;
 			}
@@ -1654,7 +1684,18 @@ namespace Frida.Gadget {
 			public async void destroy_script (AgentScriptId script_id, Cancellable? cancellable) throws Error, IOError {
 				check_open ();
 
-				yield script_engine.destroy_script (script_id);
+				var eternalize = false;
+				var listen_interaction = server.config.interaction as ListenInteraction;
+				if (listen_interaction != null) {
+					eternalize = listen_interaction.eternalize == ListenInteraction.EternalizeBehavior.YES;
+				}
+
+				if (eternalize) {
+					var script = script_engine.eternalize_script (script_id);
+					script_eternalized (script);
+				} else {
+					yield script_engine.destroy_script (script_id);
+				}
 			}
 
 			public async void load_script (AgentScriptId script_id, Cancellable? cancellable) throws Error, IOError {


oleavr:
I realize this isn't covered by our docs, but Gadget does support `spawn("re.frida.Gadget")`. This makes it possible to do early instrumentation the usual way – calling it sets a flag so that the following `attach()` doesn't implicitly `resume()`. It means existing tools can support regular vs gadget without any knowledge of what they're talking to.

Basically:
```sh
$ frida -U -f re.frida.Gadget --no-pause -l explore.js
```

Will load `explore.js`, and then call `resume()`.

The `frida-inject` tool supports eternalizing, but frida-tools don't yet. (That would be good to support though.)

oleavr:
By the way, there is no need to embed the gadget on iOS for `listen` use-cases – you can `frida-ps -Uai` and spawn any debuggable app, and Frida will inject gadget dynamically. (No jailbreak required.)

ChaosData:
Ok, `-f re.frida.Gadget` and `%resume` (which didn't work from non-spawned mode and only works with the repl due to it being a magic command instead of a JS API) are somewhat useful in my case, though without the `--no-pause`. It does still run into issues with the process exiting before `console.log()` can emit back to the client though (but throwing in `Thread.sleep()` can help). However, it's definitely the case that using `-q -l script.js` results in losing the hooks due to lack of eternalizing since the client will disconnect. Is there any interest in exposing some of the DBus-only APIs to the interpreter itself (or enabling an interpreter-run client)?

I'm currently working with a non-standard Linux system (ChromeOS) where Frida's injection mode doesn't work (this is known, but fwiw I've been adding a ton of logging to a custom build to try and track down where the flow is breaking down), so I've been making do with the gadget in the meantime. As an aside, do you know of any particular reason the `enable_child_gating()` logic from https://github.com/frida/frida-core/blob/b22fb70/lib/agent/agent.vala#L734 wouldn't work in the gadget? In the meantime I'm currently exploring methods of injecting the gadget.so into processes directly and seeing if I can hackily force the already in-memory gadget.so post fork to be executed with or without using ptrace.

oleavr:
> Is there any interest in exposing some of the DBus-only APIs to the interpreter itself

It's already exposed to the Python bindings. It's just the REPL that doesn't make use of it. It's only a matter of calling `script.eternalize()` instead of `script.unload()`.

> As an aside, do you know of any particular reason the enable_child_gating() logic from https://github.com/frida/frida-core/blob/b22fb70/lib/agent/agent.vala#L734 wouldn't work in the gadget?

It should be possible to implement it, but it will probably be very different given that it won't rely on an external process to help it through the `exec()` transition. But there's gotta be a way. E.g. by playing with `LD_PRELOAD` perhaps. (Haven't given this much thought yet, so any research on this is much appreciated!)

oleavr:
(Assuming this can be closed. Let me know if there's still something missing.)

PR #287, opened 2020-01-23
Fix crash of target process in windows injection

dasraf9:
On failed injection attempt a breakpoint call causes some target
processes to crash.

Add a call to TerminateThread instead for a graceful exit.

### DIFF
--- a/src/windows/winjector-helper-core.c
+++ b/src/windows/winjector-helper-core.c
@@ -49,6 +49,7 @@ struct _RemoteWorkerContext
   gpointer get_proc_address_impl;
   gpointer free_library_impl;
   gpointer virtual_free_impl;
+  gpointer get_last_error_impl;
 
   WCHAR dll_path[MAX_PATH + 1];
   gchar entrypoint_name[256];
@@ -322,6 +323,7 @@ initialize_remote_worker_context (RemoteWorkerContext * rwc,
   const gsize data_alignment = 4;
   const gchar * loadlibrary_failed_label = "loadlibrary_failed";
   const gchar * skip_unload_label = "skip_unload";
+  const gchar * restore_registers = "restore_registers";
 
   gum_init ();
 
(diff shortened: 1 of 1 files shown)


src/windows/winjector-helper-core.c
```diff
@@ -380,18 +382,27 @@ initialize_remote_worker_context (RemoteWorkerContext * rwc,
   /* } */
   gum_x86_writer_put_label (&cw, skip_unload_label);
 
+  /* set return 0 */
+  gum_x86_writer_put_xor_reg_reg(&cw, GUM_REG_EAX, GUM_REG_EAX);
```
oleavr: Inconsistent style as per:
https://github.com/frida/frida/blob/master/CONTRIBUTING.md#rule-12

(Here and elsewhere, including macro invocations.)

src/windows/winjector-helper-core.c
```diff
@@ -380,18 +382,27 @@ initialize_remote_worker_context (RemoteWorkerContext * rwc,
...
+
+  /* if (call GetLastError) { */
+  gum_x86_writer_put_jcc_short_label(&cw, X86_INS_JE, restore_registers, GUM_NO_HINT);
```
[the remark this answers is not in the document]
dasraf9: It should be conditional because we don't know what GetLastError will return in a normal case. When everything went good we don't want to enter this block. The xor is always setting zf to 0 so we won't enter this block is a normal case.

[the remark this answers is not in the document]
oleavr: But there is no other path to this instruction than the xor, so why isn't it an unconditional jump?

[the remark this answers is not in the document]
dasraf9: You are correct, I'll change it.

src/windows/winjector-helper-core.c
```diff
@@ -380,18 +382,27 @@ initialize_remote_worker_context (RemoteWorkerContext * rwc,
...
+  gum_x86_writer_put_label(&cw, loadlibrary_failed_label);
+  gum_x86_writer_put_call_reg_offset_ptr_with_arguments (&cw, GUM_CALL_SYSAPI,
+      GUM_REG_XBX, G_STRUCT_OFFSET(RemoteWorkerContext, get_last_error_impl),
+      1,
```
oleavr: This should be 0 and the next two arguments omitted. (`GetLastError()` doesn't expect any arguments.)

dasraf9: done.

src/windows/winjector-helper-core.c
dasraf9: done.

oleavr:
(Sorry for the delay here, currently heads down in iOS land; will make the final tweaks and land this before the upcoming release.)

oleavr:
Landed in 4320cb01be5ae763fd893dc821fafe8e96a4b030. Thanks! 🙌

PR #291, opened 2020-03-09
Add support for Unc0ver 4.3.1

mrmacete:
- use the `substitute_setup_process` mig call to `substituted` to enable early instrumentation
- ensure `substitute-inserter.dylib` is initialized right after attaching

Thanks to @pwn20wndstuff for the kind help!


src/darwin/policy-softener-glue.c
```diff
@@ -33,4 +34,42 @@ _frida_electra_policy_softener_internal_jb_entitle_now (void * jbd_call, guint s
...
   return jbd_call_func (service_port, 1, pid);
 }
+
+guint
+_frida_unc0ver_policy_softener_internal_connect (void)
+{
+  mach_port_t service_port = MACH_PORT_NULL;
+  kern_return_t kr;
+
+  kr = task_get_special_port (mach_task_self(), TASK_SEATBELT_PORT, &service_port);
```
oleavr:
```suggestion
  kr = task_get_special_port (mach_task_self (), TASK_SEATBELT_PORT, &service_port);
```

src/darwin/frida-helper-backend.vala
```diff
@@ -53,6 +53,8 @@ namespace Frida {
 #if IOS
 			if (ElectraPolicySoftener.is_available ())
 				policy_softener = new ElectraPolicySoftener ();
+			else if (Unc0verPolicySoftener.is_available())
```
oleavr:
```suggestion
			else if (Unc0verPolicySoftener.is_available ())
```

src/darwin/policy-softener-glue.c
```diff
@@ -33,4 +34,42 @@ _frida_electra_policy_softener_internal_jb_entitle_now (void * jbd_call, guint s
...
+  if (kr != KERN_SUCCESS)
+    return MACH_PORT_NULL;
+
+  if (!MACH_PORT_VALID (service_port))
```
oleavr: Theoretically if this check fails because `service_port` is `MACH_PORT_DEAD`, will we be leaking a dead port right? (Can't remember the Mach semantics about those.)

mrmacete: hmm i'm not sure this can really be `MACH_PORT_DEAD`, actually.

but maybe we could just ignore the check and let it fail (silently) later?

oleavr: Sounds good to me!

pwn20wndstuff:
I reviewed the code prior to the pull request. It is safe to merge.

pwn20wndstuff:
You shouldn’t remove the MACH_PORT_VALID check. The seatbelt special port exists by default on iOS but it will return MACH_PORT_NULL. You can’t leak a dead port name.

PR #300, opened 2020-04-07
Sign helpers callbacks

mrmacete:
This makes jailed early instrumentation work on arm64e processes, even when the legacy code path is taken.

### DIFF
--- a/src/fruity/injector.vala
+++ b/src/fruity/injector.vala
@@ -779,15 +779,15 @@ namespace Frida.Fruity.Injector {
 
 			yield save_main_thread_state (cancellable);
 
-			var invalidAsNop = new InvalidAsNopHandler();
+			var invalid_as_nop = new InvalidAsNopHandler ();
 
 			uint64 signed_ret_gadget = yield invoke_remote_function (sign_pointer, {
 					ret_gadget
-				}, invalidAsNop, cancellable);
+				}, invalid_as_nop, cancellable);
 
 			uint64 signed_get_thread_buf = yield invoke_remote_function (sign_pointer, {
 					get_thread_buf
-				}, invalidAsNop, cancellable);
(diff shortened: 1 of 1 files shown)


mrmacete:
ahh right, i was wrongly assuming that `PACIZA` acted like a NOP on non-arm64e...

mrmacete:
> we might need this logic regardless of whether pointer authentication is used for the current process
what do you mean? in which case it should be needed if the current process is not arm64e?

mrmacete:
but yeah in the worst case i think we could make `invoke_remote_function` treat invalid instructions as NOPs...

oleavr:
> ahh right, i was wrongly assuming that PACIZA acted like a NOP on non-arm64e...

That may be, I was thinking about what will happen on older CPUs that don't support the instruction.

> what do you mean? in which case it should be needed if the current process is not arm64e?

The dynamic linker is the same for both arm64e and arm64 processes, but we just have to make sure that signing the pointer in the arm64 case doesn't cause any issues.

> but yeah in the worst case i think we could make invoke_remote_function treat invalid instructions as NOPs...

Yeah or we can try/catch locally and special-case it that way. (In case "invalid instruction" can happen in initializers when an assertion fails.)

src/fruity/injector.vala
```diff
@@ -768,12 +777,24 @@ namespace Frida.Fruity.Injector {
 
 			size_t error_buf_literal_offset = code_builder.skip (4).offset;
 
+			yield save_main_thread_state (cancellable);
+
+			var invalidAsNop = new InvalidAsNopHandler();
```
oleavr, line 782:
```suggestion
			var invalid_as_nop = new InvalidAsNopHandler ();
```

oleavr:
Thanks! 💥🙌

PR #304, opened 2020-04-16
Changes to fix multiple definition error when using musl

WorksButNotTested:


WorksButNotTested:
I had previously been working with just frida-gum and injecting the frida-gadget.so using my own tools or via LD_PRELOAD, now that I have hopefully addressed the issues I was finding with frida-inject (both related to the auxilliary vector) I am hoping to move to just using FRIDA out of the box.

Hence I have now tried building the core with musl and uclibc. Then I stumbled across this really weird issue. Hope the fix is ok.

lib/payload/cloak.vala
```diff
@@ -99,23 +99,32 @@ namespace Frida {
 			listeners.add (readdir_listener);
 			interceptor.attach (readdir, readdir_listener);
 
-			var readdir64 = Gum.Module.find_export_by_name (libc, "readdir64");
-			if (readdir64 != null && readdir64 != readdir) {
+			/*
+			 * It seems that when using the musl CRT, the compiler incorrectly
+			 * determines that symbols with the same name, but a different
```
oleavr, line 104: Hmm, perhaps the reason is that it has `#define readdir64` to something, and that's getting picked up here?

lib/payload/cloak.vala
```diff
@@ -99,23 +99,32 @@ namespace Frida {
...
+			 * numeric suffix are the same symbol. Thus when the C code is
+			 * generate for this VALA, the symbols 'readdir' and 'readdir64'
+			 * collide resulting in a redefinition error. Hence the name has
+			 * been slightly changed.
+			 */
+			var readdir64_w = Gum.Module.find_export_by_name (libc,
+			"readdir64");
```
oleavr, line 111: Lines should not be broken at 80 cols, but 140 cols in frida-core.

lib/payload/cloak.vala
```diff
@@ -99,23 +99,32 @@ namespace Frida {
...
 			interceptor.attach (readdir_r, readdir_r_listener);
 
-			var readdir64_r = Gum.Module.find_export_by_name (libc, "readdir64_r");
-			if (readdir64_r != null && readdir64_r != readdir_r) {
+			var readdir64_r_w = Gum.Module.find_export_by_name (libc, "readdir64_r");
+			if (readdir64_r_w != null && readdir64_r_w != readdir_r) {
 				var listener = new ReadDirRListener (this, MODERN);
 				listeners.add (listener);
-				interceptor.attach (readdir64_r, listener);
+				interceptor.attach (readdir64_r_w, listener);
```
oleavr, line 127: Let's use `_impl` as suffix instead.

lib/payload/cloak.vala
WorksButNotTested: Oh maybe, if they had something like:
#define readdir64 readdir

Then that could throw it off. Then the compiler warning message was showing the source file before it had been pre-processed and hence showing the name collision. I have updated the comment accordingly and fixed it to wrap at 140 cols.

lib/payload/cloak.vala
```diff
@@ -99,23 +99,27 @@ namespace Frida {
...
 
-			var readdir64_r = Gum.Module.find_export_by_name (libc, "readdir64_r");
-			if (readdir64_r != null && readdir64_r != readdir_r) {
+			var readdir64_r_impl = Gum.Module.find_export_by_name (libc, "readdir64_r");
+			if (readdir64_r_impl != null && readdir64_r_impl != readdir_r) {
 				var listener = new ReadDirRListener (this, MODERN);
 				listeners.add (listener);
-				interceptor.attach (readdir64_r, listener);
+				interceptor.attach (readdir64_r_impl, listener);
 			}
```
oleavr, line 123: I think we should also suffix the other names for consistency. I'm not sure we need the comment though, as it's generally a bad idea to name variables the same as standard functions.

WorksButNotTested:
Made names consistent and removed comment.

PR #307, opened 2020-04-19
Add read write mmap munmap libc-shims 64-bit darwin

mrmacete:


lib/payload/libc-shim.c
```diff
@@ -423,3 +423,204 @@ _frida_set_errno (int n)
...
+
+#if defined (HAVE_DARWIN) && GLIB_SIZEOF_VOID_P == 8
+
+# undef read
+# undef write
+# undef mmap
+# undef munmap
+
+ssize_t
+read(int fildes, const void *buf, size_t nbyte)
```
oleavr, line 435:
```suggestion
read (int fildes, const void * buf, size_t nbyte)
```

lib/payload/libc-shim.c
```diff
@@ -423,3 +423,204 @@ _frida_set_errno (int n)
...
+        "r" (nbyte)
+      : "x0", "x1", "x2", "x16"
+  );
+# endif
+
+  return result;
+}
+
+ssize_t
+write(int fildes, const void *buf, size_t nbyte)
```
oleavr, line 480:
```suggestion
write (int fildes, const void * buf, size_t nbyte)
```

lib/payload/libc-shim.c
```diff
@@ -423,3 +423,204 @@ _frida_set_errno (int n)
...
+        "r" (nbyte)
+      : "x0", "x1", "x2", "x16"
+  );
+# endif
+
+  return result;
+}
+
+void *
+mmap(void *addr, size_t len, int prot, int flags, int fd, off_t offset)
```
oleavr, line 525:
```suggestion
mmap (void * addr, size_t len, int prot, int flags, int fd, off_t offset)
```

lib/payload/libc-shim.c
```diff
@@ -423,3 +423,204 @@ _frida_set_errno (int n)
...
+        "r" (offset)
+      : "x0", "x1", "x2", "x3", "x4", "x5", "x16"
+  );
+# endif
+
+  return result;
+}
+
+int
+munmap(void *addr, size_t len)
```
oleavr, line 586:
```suggestion
munmap (void * addr, size_t len)
```

PR #312, opened 2020-05-01
Relaxed post restriction in LOADING state

Gbps:
Scripts in the LOADING state currently are unable to have messages posted to them. This makes scripts unable to use blocking `recv()` while in their main body. This restriction is unnecessary as LOADING scripts are perfectly capable of receiving messages.

Tested and working with blocking `recv()` in local build.

Fix to issue: https://github.com/frida/frida-gum/issues/420

### DIFF
--- a/lib/payload/script-engine.vala
+++ b/lib/payload/script-engine.vala
@@ -436,7 +436,7 @@ namespace Frida {
 					case LOADING:
 					case LOADED:
 					case DISPOSED:
-					throw new Error.INVALID_OPERATION ("Only scripts still alive may be posted to");
+						throw new Error.INVALID_OPERATION ("Only active scripts may be posted to");
 					default:
 						script.post (message, data);
 						break;
--- a/lib/payload/script-engine.vala
+++ b/lib/payload/script-engine.vala
@@ -432,10 +432,15 @@ namespace Frida {
 			}
 
 			public void post (string message, Bytes? data) throws Error {
-				if (state != LOADED && state != DISPOSED && state != LOADING)
-					throw new Error.INVALID_OPERATION ("Only loading/loaded scripts may be posted to");
-
-				script.post (message, data);
+				switch(state) {
+					case LOADING:
+					case LOADED:
+					case DISPOSED:
+					throw new Error.INVALID_OPERATION ("Only scripts still alive may be posted to");
+					default:
+					script.post (message, data);
+					break;
+				}
 			}
 
 			private void on_message (Gum.Script script, string raw_message, Bytes? data) {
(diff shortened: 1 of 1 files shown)


lib/payload/script-engine.vala
```diff
@@ -432,8 +432,8 @@ namespace Frida {
 			}
 
 			public void post (string message, Bytes? data) throws Error {
-				if (state != LOADED && state != DISPOSED)
-					throw new Error.INVALID_OPERATION ("Only loaded scripts may be posted to");
+				if (state != LOADED && state != DISPOSED && state != LOADING)
+					throw new Error.INVALID_OPERATION ("Only loading/loaded scripts may be posted to");
```
oleavr, line 436: Would be good to turn this into a switch statement, and order the cases as in the enum. Perhaps we should also simplify the error message, as it's already misleading (we allow it in `DISPOSED` state too). Something along the lines of: `Only scripts still alive may be posted to`

Gbps:
Reformatted. Something like this?

lib/payload/script-engine.vala
```diff
@@ -432,10 +432,15 @@ namespace Frida {
 			}
 
 			public void post (string message, Bytes? data) throws Error {
-				if (state != LOADED && state != DISPOSED && state != LOADING)
-					throw new Error.INVALID_OPERATION ("Only loading/loaded scripts may be posted to");
-
-				script.post (message, data);
+				switch(state) {
```
oleavr, line 435:
```suggestion
				switch (state) {
```

lib/payload/script-engine.vala
```diff
@@ -432,10 +432,15 @@ namespace Frida {
...
+					case LOADING:
+					case LOADED:
+					case DISPOSED:
+					throw new Error.INVALID_OPERATION ("Only scripts still alive may be posted to");
```
oleavr, line 439:
```suggestion
						throw new Error.INVALID_OPERATION ("Only active scripts may be posted to");
```

(Fixing indent and making it clearer for the case where somebody forgets to call `.load ()`.)

lib/payload/script-engine.vala
```diff
@@ -432,10 +432,15 @@ namespace Frida {
...
+					default:
+					script.post (message, data);
+					break;
```
oleavr, line 442:
```suggestion
						script.post (message, data);
						break;
```

lib/payload/script-engine.vala
oleavr: Or perhaps: `Only live scripts may be posted to`

oleavr:
Thanks a lot! 🚀

oleavr:
Oops, this inverted the logic and I totally missed that. 😅 Just pushed a [fix](https://github.com/frida/frida-core/commit/a06aef06b02360a825ae7a94a1626dc52b4a24de). 🙌

PR #313, opened 2020-05-05
Added the ability to run frida-server on a non-root Android device.

KenjiroIchise:
I've made some fixes to make frida-server work on non-root terminals.
If a file named "fridakeep" exists in the same directory as frida-server, the following is applied.

1.disabling selinux checks
2.On Android, change the default "/data/local/tmp" to the path described in fridakeep.
```
/data/data/com.app.name
```
As with frida-gadget, some advance preparation is required.
1.APK:AndroidManifest.xml in android:debuggable="true"
2.command
```
$ adb shell
$ pm list packages
$ run-as com.app.name
$ cp /data/local/tmp/frida-server ./frida-server
$ ./frida-server
```
I hope that this fix will contribute to the development of security diagnostics.

### DIFF
--- a/lib/pipe/pipe-unix.c
+++ b/lib/pipe/pipe-unix.c
@@ -27,7 +27,23 @@ frida_pipe_transport_get_temp_directory (void)
   if (temp_directory != NULL)
     return temp_directory;
   else
+  {
+    char buf[512];
+    FILE *fp=NULL;
+    char *bufp;
+    
+    fp = fopen("fridakeep","r");
+    if(fp)
+    {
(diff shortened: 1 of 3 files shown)


KenjiroIchise:
I've noticed that the current implementation causes errors in at least the following features
frida-trace:only active scripts may be posted to
process-spawn:unable to find process with name 'system_server'

--Hooking up from a script from a python binding or something like that was no problem.

Is it possible to determine whether the above error can't be done theoretically or whether it is a problem that can be avoided by implementation?
And do you think that it is meaningful to run frida-server by such a method in the first place?

PR #315, opened 2020-05-13
Remove tcp check in android open_channel so that other protocols are supported

aemmitt-ns:
This small change just removes the check for the "tcp:" prefix in open_channel so that other protocols like localabstract, localreserved, etc will work. Droidy.Client.request_protocol_change already handles this appropriately. Now the only check is that the address has the format "%s:%s".

### DIFF
--- a/src/droidy/droidy-host-session.vala
+++ b/src/droidy/droidy-host-session.vala
@@ -230,17 +230,16 @@ namespace Frida {
 		}
 
 		public async IOStream open_channel (string address, Cancellable? cancellable = null) throws Error, IOError {
-			if (address.has_prefix ("tcp:")) {
-				ulong raw_port;
-				if (!ulong.try_parse (address.substring (4), out raw_port) || raw_port == 0 || raw_port > uint16.MAX)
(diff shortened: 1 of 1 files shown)


src/droidy/droidy-host-session.vala
```diff
@@ -230,17 +230,12 @@ namespace Frida {
 		}
 
 		public async IOStream open_channel (string address, Cancellable? cancellable = null) throws Error, IOError {
-			if (address.has_prefix ("tcp:")) {
-				ulong raw_port;
-				if (!ulong.try_parse (address.substring (4), out raw_port) || raw_port == 0 || raw_port > uint16.MAX)
-					throw new Error.INVALID_ARGUMENT ("Invalid TCP port");
-				uint16 port = (uint16) raw_port;
-
+			if (address.split(":").length == 2) {
```
oleavr, line 233:
```suggestion
			if (address.contains (":")) {
```

- Missing space.
- Only check that there's at least one ":" as the name of an abstract UNIX domain socket could theoretically contain ":".

PR #327, opened 2020-07-07
Aux option to specify Android user during app spawn

sowdust:
Although seldom used, Android can be a multi user environment. For example, BYOD enterprise applications often use "Android work profiles" to keep personal and work environments separate; this works by leveraging the multi user features. Standard Android APIs by default work with the default user (0). By using a different set of APIs, this commit allows to select a specific user id to spawn the application. It can be passed as an "aux" option, so it can be combined with other android specific options such as "activity".

### DIFF
--- a/src/darwin/frida-helper-process.vala
+++ b/src/darwin/frida-helper-process.vala
@@ -437,8 +437,8 @@ namespace Frida {
 					var size = input.read_uint32 ();
 					input.skip (4);
 
-					bool is_arm64 = cputype == 0x100000c;
-					bool is_arm64e = is_arm64 && (cpusubtype & 0x00ffffff) == 2;
+					bool is_arm64 = cputype == 0x0100000cU;
+					bool is_arm64e = is_arm64 && (cpusubtype & 0x00ffffffU) == 2;
 					if (is_arm64e) {
 						arm64e_offset = offset;
 					} else if (is_arm64) {


src/linux/agent/system-server.js
```diff
@@ -26,7 +26,7 @@ function init() {
 }
 
 rpc.exports = {
-  enumerateApplications: function () {
+  enumerateApplications: function (userid=0) {
```
oleavr, line 29: Should only use ES5 syntax here so Duktape can load this agent in case Frida is built without V8.

src/linux/agent/system-server.js
```diff
@@ -75,19 +76,29 @@ rpc.exports = {
...
+      if(userid != 0) {
+        var UserHandle = Java.use('android.os.UserHandle');
+        var userHandle = UserHandle.of(userid);
+        context = context.createPackageContextAsUser(pkg, GET_META_DATA, userHandle);
+        packageManager = context.getPackageManager();
+      }
+      
       var appInstalled = false;
-      var apps = packageManager.getInstalledApplications(GET_META_DATA);
+      var apps = packageManager.getInstalledApplicationsAsUser(GET_META_DATA, userid);
```
oleavr, line 101: Trailing whitespace.

src/linux/agent/system-server.js
```diff
@@ -75,19 +76,29 @@ rpc.exports = {
...
-        throw new Error("Unable to find application with identifier '" + pkg + "'");
+        throw new Error("Unable to find application with identifier '" + pkg + "' for user " + userid);
       }
     });
   },
-  startActivity: function (pkg, activity) {
+  startActivity: function (pkg, activity, userid=0) {
     return performOnJavaVM(function () {
+
+      if(userid != 0) {
```
oleavr, line 93:
```suggestion
      if (uid !== 0) {
```

(either that or `userId`)

src/linux/linux-host-session.vala
```diff
@@ -688,28 +689,38 @@ namespace Frida {
...
 			var process_name = yield call ("getProcessName", new Json.Node[] { package_name_value }, cancellable);
 
 			return process_name.get_string ();
 		}
 
+		public async string get_process_name_for_user (string package, uint userid, Cancellable? cancellable) throws Error, IOError {
+			var package_name_value = new Json.Node.alloc ().init_string (package);
+			var userid_value = new Json.Node.alloc ().init_int (userid);
+			var process_name = yield call ("getProcessName", new Json.Node[] { package_name_value, userid_value }, cancellable);
+
```
oleavr, line 701:
```suggestion
```

src/linux/linux-host-session.vala
```diff
@@ -1177,6 +1189,9 @@ namespace Frida {
 	}
 
 	private class PackageEntrypoint : Object {
+
```
oleavr, line 1192:
```suggestion

```

src/linux/linux-host-session.vala
```diff
@@ -1219,6 +1234,13 @@ namespace Frida {
 			if (entrypoint == null)
 				entrypoint = new DefaultActivityEntrypoint ();
 
+			if (aux_options.contains ("userid")) {
```
oleavr, line 1237:
```suggestion
			if (aux_options.contains ("uid")) {
```

src/linux/linux-host-session.vala
```diff
@@ -1219,6 +1234,13 @@ namespace Frida {
 			if (entrypoint == null)
 				entrypoint = new DefaultActivityEntrypoint ();
 
+			if (aux_options.contains ("userid")) {
+				string? string_userid = null;
+				if (!aux_options.lookup ("userid", "s", out string_userid))
```
oleavr, line 1239: Here we should be expecting a uint, not a string.

sowdust: Honestly, I tried several ways but I could not manage to address this bit. 
Using any numeric format string always fails, also when manually setting the type via python

oleavr: Out of all of the types supported by [GVariant](https://developer.gnome.org/glib/stable/gvariant-format-strings.html), I elected to only support three types: `string`, `int64`, and `boolean`. We could expand on this in the future, as long as we try to keep it to a minimum in order to ease the burden on each language binding.

Something like this should work:

```
				int64 uid = 0;
				if (!aux_options.lookup ("uid", "x", out uid))
```

(And then cast that to `uint` afterwards.)

src/linux/agent/system-server.js
sowdust: addressed by 5fbd9a441df529369caa3bac55fead2cf65d1892

src/linux/agent/system-server.js
sowdust: 2d966a86f9027d314d2ecff7e38c9ff2d95ae0c7

src/linux/agent/system-server.js
sowdust: 2d966a86f9027d314d2ecff7e38c9ff2d95ae0c7

src/linux/linux-host-session.vala
sowdust: 115186c914b0cdfd3f4eced978cbff8502c5a5e3

src/linux/linux-host-session.vala
sowdust: 115186c914b0cdfd3f4eced978cbff8502c5a5e3

src/linux/linux-host-session.vala
sowdust: 4cfb769fdf2ccc55186d2644cbb1ab7db70a5c73

src/linux/linux-host-session.vala
sowdust: Oh ok, thanks very much for the explanation. 
Addressed by abf41cf2b57c1af01cb400b4a6ff3a986e3bfb8d

oleavr:
Oops, I cloned this locally and pushed a tweaked version to your branch – but then realized I pushed the wrong local branch 🤦 Sorry about that. I'll just take it for a spin and then push it to master.

oleavr:
Landed as 1560d29710d9b48b11c1349815eb0a06a35fd422. Thanks! 🙌

PR #328, opened 2020-07-08
Daemonize helper process on Darwin

mrmacete:
In this way there’s no need to install a `SIGCHLD` handler which could conflict with others already installed in the process, causing issues for example with NodeJS and its `child_process` logic.

### DIFF
--- a/src/darwin/frida-helper-process.vala
+++ b/src/darwin/frida-helper-process.vala
@@ -2,10 +2,7 @@ namespace Frida {
 	public class DarwinHelperProcess : Object, DarwinHelper {
 		public uint pid {
 			get {
-				if (process == null)
-					return 0;
-
-				return (uint) uint64.parse (process.get_identifier ());
+				return process_pid;
 			}
 		}
 
@@ -16,8 +13,8 @@ namespace Frida {
 
 		private ResourceStore _resource_store;
 
+		private uint process_pid;
 		private MainContext main_context;
-		private Subprocess process;
 		private TaskPort task;
 		private DBusConnection connection;
 		private DarwinRemoteHelper proxy;
@@ -50,7 +47,7 @@ namespace Frida {
 				}
 			}
 
-			process = null;
+			process_pid = 0;
 
 			_resource_store = null;
 		}
@@ -255,7 +252,7 @@ namespace Frida {
 			}
 			obtain_request = new Promise<DarwinRemoteHelper> ();
 
-			Subprocess pending_process = null;
+			uint peer_pid = 0;
 			TaskPort pending_task_port = null;
 			DBusConnection pending_connection = null;
 			DarwinRemoteHelper pending_proxy = null;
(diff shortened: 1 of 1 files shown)


src/darwin/frida-helper-process.vala
```diff
@@ -267,9 +264,10 @@ namespace Frida {
 				var handshake_port = new HandshakePort.local (service_name);
 
 				string[] argv = { get_resource_store ().helper.path, service_name };
-				pending_process = new Subprocess.newv (argv, SubprocessFlags.INHERIT_FDS);
 
-				var peer_pid = (uint) uint64.parse (pending_process.get_identifier ());
+				GLib.SpawnFlags flags = GLib.SpawnFlags.LEAVE_DESCRIPTORS_OPEN | /*GLib.SpawnFlags.CLOEXEC_PIPES*/ 256;
```
mrmacete, line 268: I don't know why, but apparently `GLib.SpawnFlags.CLOEXEC_PIPES` doesn't seem to be exposed (even if it's documented) and i get an error like:
```
../../../frida-core/src/darwin/frida-helper-process.vala:266.70-266.98: error: The name `CLOEXEC_PIPES' does not exist in the context of `GLib.SpawnFlags' (glib-2.0)
				GLib.SpawnFlags flags = GLib.SpawnFlags.LEAVE_DESCRIPTORS_OPEN | GLib.SpawnFlags.CLOEXEC_PIPES;
				                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

mrmacete: i'm not even sure we really need it, i kept it for compatibility with what SubProcess was already doing here.

oleavr, line 268: Makes sense!  Yeah this looks like a vapi binding bug.

```suggestion
				GLib.SpawnFlags flags = GLib.SpawnFlags.LEAVE_DESCRIPTORS_OPEN | /* GLib.SpawnFlags.CLOEXEC_PIPES */ 256;
```

src/darwin/frida-helper-process.vala
```diff
@@ -16,8 +13,8 @@ namespace Frida {
 
 		private ResourceStore _resource_store;
 
+		private uint process_pid;
```
oleavr, line 16: Let's move this down to where `process` was.

src/darwin/frida-helper-process.vala
```diff
@@ -255,7 +252,7 @@ namespace Frida {
 			}
 			obtain_request = new Promise<DarwinRemoteHelper> ();
 
-			Subprocess pending_process = null;
+			uint peer_pid = 0;
```
oleavr, line 255:
```suggestion
			uint pending_pid = 0;
```

(To make it clear it belongs with the other ones.)

src/darwin/frida-helper-process.vala
```diff
@@ -302,8 +300,9 @@ namespace Frida {
 				obtain_request.resolve (proxy);
 				return proxy;
 			} else {
-				if (pending_process != null)
-					pending_process.force_exit ();
+				process_pid = 0;
```
oleavr, line 303: Shouldn't need to clear it here as we haven't set it.

mrmacete:
I couldn't reproduce the error in the same way in Linux. Since it's using the same APIs it's possible that the problem is there as well, just i'm not sure how to reproduce it systematically.

I could do the same tweak also there but i'm a bit blind for validating it.

oleavr:
> I could do the same tweak also there but i'm a bit blind for validating it.

That would be great! I've seen random test failures on Linux that look a lot like the same underlying issue. It's very racy though, so not easy to verify.

PR #329, opened 2020-07-11
Fix initial device list logic on fruity host session

mrmacete:
usbmuxd doesn't accept any command after starting the listening mode. The “dummy request” we used to flush the device list makes usbmuxd [close the connection](https://github.com/libimobiledevice/usbmuxd/blob/master/src/client.c#L655) because the command is received in the wrong state.

To avoid this while ensuring the initial device list is as close as possible to the actual set of connected devices, the new logic is:
- call ListDevices
- start listening for changes in the list
- try to connect to all devices in the initial list, to
ensure they’ve not been detached in the meantime
- remove detached devices from the list

### DIFF
--- a/src/fruity/fruity-host-session.vala
+++ b/src/fruity/fruity-host-session.vala
@@ -47,6 +47,7 @@ namespace Frida {
 		private async void do_start () {
 			bool success = true;
 			Fruity.DeviceId[] initial_device_ids = {};
+			Fruity.UsbmuxClient list_client = null;
 
 			try {
 				control_client = yield Fruity.UsbmuxClient.open (start_cancellable);
@@ -58,10 +59,16 @@ namespace Frida {
 					remove_device (id);
 				});
 
-				initial_device_ids = yield control_client.list_devices (start_cancellable);
 				yield control_client.enable_listen_mode (start_cancellable);
+
+				list_client = yield Fruity.UsbmuxClient.open (start_cancellable);
+				initial_device_ids = yield list_client.list_devices (start_cancellable);
 			} catch (GLib.Error e) {
 				success = false;
+			}
+
+			if (list_client != null) {
+				list_client.close.begin (null);
 			}
 
 			if (success) {
@@ -78,7 +85,7 @@ namespace Frida {
 							remove_device (id);
 						}
 
-						yield validate_client.close (null);
+						validate_client.close.begin (null);
 					}
 				} catch (GLib.Error e) {
 					success = false;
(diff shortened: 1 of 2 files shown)


src/fruity/fruity-host-session.vala
```diff
@@ -57,19 +58,34 @@ namespace Frida {
 					remove_device (id);
 				});
 
+				initial_device_ids = yield control_client.list_devices (start_cancellable);
```
oleavr, line 61: Doesn't this still leave a tiny window where a new device could be added and we'd miss it?

mrmacete: possibly. would it help to start listening first and get the list from another connection?

oleavr: Yeah I think that should eliminate it!

oleavr: Actually I just thought of another option: if the daemon is single-threaded, we can actually just do a `ListDevices` right after `Listen` has completed. Once that `ListDevices` has completed we are done, and we can ignore the list of devices that it responded with.

mrmacete: i have to check, i was assuming one thread pre client, but i've got to read some code (or just place logs) in usbmuxd

mrmacete:
Yeah this solution is slower than before. With 2 devices the new implementation takes around 4 milliseconds, while the previous implementation took less than 2 milliseconds. so it's roughly 2 times slower (as measured by vala Timer class)

oleavr:
Ah cool! That's not too bad.

src/fruity/fruity-host-session.vala
mrmacete: LOL ok yeah usbmuxd has a single thread!
so in theory we can use the previous implementation, just on 2 separate connections 🤦

oleavr: Ah, interesting! I wouldn't be surprised if Apple's implementation is way more advanced though. But if that's the case I'm not sure if we can fingerprint the daemon very easily – perhaps there's a command we can use?

mrmacete: yeah the only way i know to tell the implementations apart is maybe to trigger this original issue in a controlled way, like:

- do what we always did (which works in Apple's implementation)
- detect the underlying connection has been closed in usbmuxd case
- redo everything using 2 different connections in case the connection has been closed

(not sure how to detect that, though, but i bet it's possible)

the disadvantage of that is in the usbmuxd case will still take twice longer, but maybe the code is simpler

oleavr:
Superseded by #330.

PR #330, opened 2020-07-12
Handle control_client closed by usbmuxd

mrmacete:
This is an alternative simpler solution for the same problem https://github.com/frida/frida-core/pull/329 was attempting to solve.

usbmuxd doesn’t accept any command after starting the listening mode. The “dummy request” we used to flush the device list makes usbmuxd close the connection because the command is received in the wrong state.

To avoid this issue, the new logic is:

- detect if the peer has closed the connection
- in that case reopen it, and send the “dummy
request” on a secondary connection

That should work because usbmuxd handles all clients on a single thread, so by the time the “dummy request” is handled, the entire list of currently connected devices should have been sent.

According to my local tests, performance-wise this is indistinguishable from the previous logic.

### DIFF
--- a/src/fruity/fruity-host-session.vala
+++ b/src/fruity/fruity-host-session.vala
@@ -48,13 +48,13 @@ namespace Frida {
 			bool success = yield try_start_control_connection ();
 
 			if (success) {
-				/* perform a dummy-request to flush out any pending device attach notifications */
+				/* Perform a dummy-request to flush out any pending device attach notifications. */
 				try {
 					yield control_client.connect_to_port (Fruity.DeviceId (uint.MAX), 0, start_cancellable);
 					assert_not_reached ();
 				} catch (GLib.Error expected_error) {
 					if (expected_error.code == IOError.CONNECTION_CLOSED) {
-						/* usbmuxd closes the connection when receiving commands in the wrong state */
+						/* Deal with usbmuxd closing the connection when receiving commands in the wrong state. */
 						control_client.close.begin (null);
 
 						success = yield try_start_control_connection ();
@@ -63,7 +63,9 @@ namespace Frida {
 							try {
 								flush_client = yield Fruity.UsbmuxClient.open (start_cancellable);
 								try {
-									yield flush_client.connect_to_port (Fruity.DeviceId (uint.MAX), 0, start_cancellable);
+									yield flush_client.connect_to_port (
+											Fruity.DeviceId (uint.MAX), 0,
+											start_cancellable);
 									assert_not_reached ();
 								} catch (GLib.Error expected_error) {
 								}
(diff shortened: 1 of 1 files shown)


src/fruity/fruity-host-session.vala
```diff
@@ -45,6 +45,51 @@ namespace Frida {
...
+								} catch (GLib.Error expected_error) {
+								}
+							} catch (GLib.Error e) {
+								success = false;
+							}
+
+							if (flush_client != null)
+								flush_client.close.begin (null);
+
+							if (success == false && control_client != null) {
```
oleavr, line 77:
```suggestion
							if (!success && control_client != null) {

```

src/fruity/fruity-host-session.vala
```diff
@@ -45,6 +45,51 @@ namespace Frida {
...
+						/* usbmuxd closes the connection when receiving commands in the wrong state */
+						control_client.close.begin (null);
+
+						success = yield try_start_control_connection ();
+						if (success) {
+							Fruity.UsbmuxClient flush_client = null;
+							try {
+								flush_client = yield Fruity.UsbmuxClient.open (start_cancellable);
+								try {
+									yield flush_client.connect_to_port (Fruity.DeviceId (uint.MAX), 0, start_cancellable);
```
oleavr, line 66: We should try to keep lines <= 140 cols. (Slowly moving frida-core towards this.)

src/fruity/fruity-host-session.vala
```diff
@@ -45,6 +45,51 @@ namespace Frida {
 		}
 
 		private async void do_start () {
+			bool success = yield try_start_control_connection ();
+
+			if (success) {
+				/* perform a dummy-request to flush out any pending device attach notifications */
```
oleavr, line 51:
```suggestion
				/* Perform a dummy-request to flush out any pending device attach notifications. */

```

(Correcting myself here. :-) )

src/fruity/fruity-host-session.vala
```diff
@@ -45,6 +45,51 @@ namespace Frida {
...
+			bool success = yield try_start_control_connection ();
+
+			if (success) {
+				/* perform a dummy-request to flush out any pending device attach notifications */
+				try {
+					yield control_client.connect_to_port (Fruity.DeviceId (uint.MAX), 0, start_cancellable);
+					assert_not_reached ();
+				} catch (GLib.Error expected_error) {
+					if (expected_error.code == IOError.CONNECTION_CLOSED) {
+						/* usbmuxd closes the connection when receiving commands in the wrong state */
```
oleavr, line 57:
```suggestion
						/* Deal with usbmuxd closing the connection when receiving commands in the wrong state. */

```

oleavr:
Thanks! 😍

PR #337, opened 2020-08-03
Fix starting Android activities when the intent is not automatically found

muhzii:
Set the flag 'FLAG_ACTIVITY_NEW_TASK' for the 'Intent' object before calling 'startActivity'. See https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/ContextImpl.java#955 for the reason why this might occasionally work when the flag is omitted.

Also remove the call to 'setAction', since explicit intents are being used.


oleavr:
Thanks!

PR #340, opened 2020-08-29
Initialize CoreFoundation on i/macOS

mrmacete:
The problem is that `CoreFoundation` must be initialised on the main thread, this means right now it's not possible to use anything depending on it during early instrumentation without crashing - unless we fall into the "legacy" dyld code path, for the presence of `DYLD_INSERT_LIBRARIES` and the second-stage jailbreak "bootstrappers" (for Chimera and Unc0ver) which also depend on CoreFoundation.

This change will call the `__CFInitialize` function if either:

- dyld takes the "modern" code path
- dyld takes the "legacy" code path and the "bootstrappers" aren't there, after loading `libSystem`

It's a bit hacky (I didn't find a way to convince dyld to initialise `CoreFoundation` early in the "modern" path), but - to my tests - on jailbreaks like checkra1n or on macOS it works way better than before.


mrmacete:
i think i could also add a test case

mrmacete:
Actually, to do that i should add a labrat with CoreFoundation - what's the easiest way to do that?

The test code is pretty simple, something like:

```javascript
function testCF () {
  const getMainPtr = Module.findExportByName(null, 'CFRunLoopGetMain');
  if (getMainPtr === null) {
    console.log('No CoreFoundation');
    return;
  }

  const getMain = new NativeFunction(getMainPtr, 'pointer', []);
  getMain();
  console.log('OK');
}
```

The behaviour without this PR's change is, when calling `getMain()`:

- on macos it hangs forever
- on iOS with checkra1n it crashes

src/darwin/frida-helper-backend-glue.m
```diff
@@ -344,6 +348,8 @@
 static void frida_spawn_instance_unset_helpers (FridaSpawnInstance * self);
 static void frida_spawn_instance_call_set_helpers (FridaSpawnInstance * self, GumDarwinUnifiedThreadState * state, mach_vm_address_t helpers);
 static void frida_spawn_instance_call_dlopen (FridaSpawnInstance * self, GumDarwinUnifiedThreadState * state, mach_vm_address_t lib_name, int mode);
+static gboolean find_cf_initialize (const GumModuleDetails * details, gpointer user_data);
```
oleavr, line 351:
```suggestion
static gboolean frida_find_cf_initialize (const GumModuleDetails * details, gpointer user_data);
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -3206,6 +3229,81 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
 #endif
 }
 
+static gboolean
+find_cf_initialize (const GumModuleDetails * details, gpointer user_data)
+{
+  FridaSpawnInstance * self = user_data;
+
+  if (strcmp (details->path, CORE_FOUNDATION) == 0)
```
oleavr, line 3237: Let's return early here so we can de-indent the following block.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -3206,6 +3229,81 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+    current_pc = state->uts.ts64.__rip;
+    state->uts.ts64.__rip = new_pc;
+
+    state->uts.ts64.__rsp -= 16;
+    write_succeeded = gum_darwin_write (self->task, state->uts.ts64.__rsp, (const guint8 *) &current_pc, sizeof (current_pc));
+    g_assert (write_succeeded);
+  }
+  else
+  {
+    guint32 temp[1];
```
oleavr, line 3277:
```suggestion
    guint32 return_address;
```

oleavr (changes requested):
w00t! 🤘

mrmacete:
So i added a test, not 100% sure that's the right way to mess with labrats, though 😅

src/darwin/frida-helper-backend-glue.m
```diff
@@ -3206,6 +3233,79 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
 #endif
 }
 
+static gboolean
+frida_find_cf_initialize (const GumModuleDetails * details, gpointer user_data)
+{
+  FridaSpawnInstance * self = user_data;
```
oleavr, line 3239:
```suggestion
  FridaSpawnInstance * self = user_data;
  GumDarwinModule * core_foundation;
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -3206,6 +3233,79 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+
+  if (strcmp (details->path, CORE_FOUNDATION) != 0)
+    return TRUE;
+
+  GumDarwinModule * core_foundation = gum_darwin_module_new_from_memory (CORE_FOUNDATION,
+      self->task, details->range->base_address,
+      GUM_DARWIN_MODULE_FLAGS_NONE, NULL);
```
oleavr, line 3246:
```suggestion
  core_foundation = gum_darwin_module_new_from_memory (CORE_FOUNDATION, self->task, details->range->base_address,
      GUM_DARWIN_MODULE_FLAGS_NONE, NULL);
```

oleavr (changes requested):

oleavr:
> So i added a test, not 100% sure that's the right way to mess with labrats, though 😅

(Sorry, I totally forgot to answer your original question about that!)  What you did was great, there's just a manual fix-up process where we for now have to edit `Makefile.macos.mk` and replace `-m32` and `-m64` with `-arch arm64` and `-arch arm64e`, rebuild the binaries, copy them to a subdir, revert them to the originals, and from the subdir run `../merge-macos-slices.sh` – ugh :-)

mrmacete:
thanks!! copying this in a not for the future!

PR #342, opened 2020-09-17
Add support for iOS 14.x secure DTX

mrmacete:
Exactly like debugserver, now it uses a secure endpoint.


oleavr:
Awesome! 🤘 Thanks! 🙌

PR #343, opened 2020-09-21
Kill via LLDB on jailed iOS

mrmacete:
To avoid killing via ProcessControl when possible, because apparently that leaves the debug server in a bad state for which killed apps sometimes appear as already running, failing early instrumentation on subsequent spawn attempts.

### DIFF
--- a/src/fruity/fruity-host-session.vala
+++ b/src/fruity/fruity-host-session.vala
@@ -806,9 +806,18 @@ namespace Frida {
 				}
 			}
 
-			var process_control = yield Fruity.ProcessControlService.open (channel_provider, cancellable);
+			try {
+				var lockdown = yield lockdown_provider.get_lockdown_client (cancellable);
+				var lldb = yield start_lldb_service (lockdown, cancellable);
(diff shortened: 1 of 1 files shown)


src/fruity/fruity-host-session.vala
```diff
@@ -806,9 +806,18 @@ namespace Frida {
...
-			var process_control = yield Fruity.ProcessControlService.open (channel_provider, cancellable);
+			try {
+				var lockdown = yield lockdown_provider.get_lockdown_client (cancellable);
+				var lldb = yield start_lldb_service (lockdown, cancellable);
+				var process = yield lldb.attach_by_pid (pid, cancellable);
 
-			yield process_control.kill (pid, cancellable);
+				lldb_session = new LLDBSession (lldb, process, null, channel_provider);
+				yield lldb_session.kill(cancellable);
+				yield lldb_session.close(cancellable);
```
oleavr, line 816:
```suggestion
				yield lldb_session.kill (cancellable);
				yield lldb_session.close (cancellable);

```

oleavr (changes requested):
Thanks! 💥

PR #344, opened 2020-10-06
Skip app extensions when searching frontmost app (on Fruity host session)

mrmacete:
Sometimes an app extension was returned as the first matched process, just for failing right after with the “Unable to resolve bundle path to bundle ID” exception.

### DIFF
--- a/src/fruity/dtx.vala
+++ b/src/fruity/dtx.vala
@@ -62,6 +62,10 @@ namespace Frida.Fruity {
 				if (process.get_optional_value ("startDate", out start_date))
 					info.start_date = start_date.to_date_time ();
 
+				NSNumber? is_application;
+				if (process.get_optional_value ("isApplication", out is_application))
+					info.is_application = is_application.boolean;
+
 				result.add (info);
 			}
 
@@ -228,6 +232,11 @@ namespace Frida.Fruity {
 		}
 
 		public bool foreground_running {
+			get;
+			set;
+		}
+
+		public bool is_application {
 			get;
 			set;
 		}
--- a/src/fruity/dtx.vala
+++ b/src/fruity/dtx.vala
@@ -53,6 +53,7 @@ namespace Frida.Fruity {
 
 				info.name = process.get_value<NSString> ("name").str;
 				info.real_app_name = process.get_value<NSString> ("realAppName").str;
+				info.is_application = process.get_value<NSNumber> ("isApplication").boolean;
 
 				NSNumber? foreground_running;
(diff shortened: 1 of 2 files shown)


src/fruity/fruity-host-session.vala
```diff
@@ -497,7 +497,8 @@ namespace Frida {
 			var device_info = yield Fruity.DeviceInfoService.open (channel_provider, cancellable);
 			var processes = yield device_info.enumerate_processes (cancellable);
 
-			var process = processes.first_match (p => p.foreground_running && p.real_app_name != SPRINGBOARD_PATH);
+			var process = processes.first_match (p => p.foreground_running && p.real_app_name != SPRINGBOARD_PATH &&
+					!p.real_app_name.contains (".appex/"));
```
oleavr, line 501: Is there some metadata we could use instead of sniffing the path? IIRC we currently don't expose all of the properties.

oleavr (commented):
w00t! 🤘

src/fruity/fruity-host-session.vala
mrmacete: interesting! checking

mrmacete: ok, i double checked and there's a promising `isApplication` field in the process object, but unfortunately that's `1` also for some of the appex (like Today widgets)

mrmacete: maybe it's worth exposing and checking it anyway to shrink the number of string comparisons to what's strictly needed

oleavr: Ah that would be great actually!  Wanna amend this one or leave it for later?

mrmacete: i'll just add here if you're fine with that

oleavr: Yep sounds great!

src/fruity/dtx.vala
```diff
@@ -62,6 +62,10 @@ namespace Frida.Fruity {
 				if (process.get_optional_value ("startDate", out start_date))
 					info.start_date = start_date.to_date_time ();
 
+				NSNumber? is_application;
+				if (process.get_optional_value ("isApplication", out is_application))
+					info.is_application = is_application.boolean;
```
oleavr, line 67: - Let's use `get_value()` here as we can assume it's already there.
- Perhaps we should move it before `foreground_running` as I guess that is a property that only applies to apps. (Let's also match the order here with that of the properties.)

PR #357, opened 2021-01-25
Generate GIR by default

meme:
- The Vala compiler is capable of emitting GIR directly, so g-ir-scanner, etc. are not required
- This is added as an install target as well


oleavr:
Thanks!

PR #358, opened 2021-02-01
frida-inject send stdin to script

WorksButNotTested:


inject/inject.vala
```diff
@@ -439,6 +474,15 @@ namespace Frida.Inject {
...
 		}
 
+		public async void call_send (string message) {
+			var msg = new Json.Node.alloc ().init_string (message);
+
+			try {
+				yield rpc_client.call ("recv", new Json.Node[] { msg }, io_cancellable);
+			} catch (GLib.Error e) {
+			}
+		}
```
oleavr, line 484: I'd put this after `call_init()` as it's called after it. Or at least we should probably guarantee that it is.

inject/inject.vala
```diff
@@ -225,6 +225,41 @@ namespace Frida.Inject {
 
 			exit_code = 0;
 
+			var fd = stdin.fileno();
```
oleavr, line 228:
```suggestion
			var fd = stdin.fileno ();
```

inject/inject.vala
```diff
@@ -225,6 +225,41 @@ namespace Frida.Inject {
...
+#if WINDOWS
+			var inchan = new IOChannel.win32_new_fd (fd);
+#else
+			var inchan = new IOChannel.unix_new (fd);
+#endif
+			inchan.add_watch (IOCondition.IN, (source, condition) => {
+				string str_return = null;
```
oleavr, line 235: I'd move this to right before `try {` so it makes it clear what's about to happen inside the `try`.

inject/inject.vala
```diff
@@ -225,6 +225,41 @@ namespace Frida.Inject {
...
+				size_t length = -1;
+				size_t terminator_pos = -1;
```
oleavr, line 237: These could be moved inside the `try` as they're unused outside it. (Actually I see now that they're nullable at the C API level, so would be better to pass `null`, assuming the vapi is correct.)

inject/inject.vala
```diff
@@ -225,6 +225,41 @@ namespace Frida.Inject {
...
+				}
+
+				try {
+					IOStatus status = inchan.read_line (out str_return, out length, out terminator_pos);
+					if (status == IOStatus.EOF) {
+						loop.quit ();
+						return false;
+					}
+
+					Idle.add(() => {
```
oleavr, line 250:
```suggestion
					Idle.add (() => {
```

inject/inject.vala
```diff
@@ -225,6 +225,41 @@ namespace Frida.Inject {
...
+						script_runner.call_send.begin (str_return);
+						return false;
+					});
```
oleavr, line 253: This should be moved outside the `try` as it doesn't throw.

oleavr (changes requested):
Nice! 🚀

inject/inject.vala
WorksButNotTested: I don't think I can move this? `inchan` is used to call `add_watch` so it needs to exist outside the callback. The try needs to handle potential errors occurring inside the callback.

oleavr: I meant the `str_return` variable, sorry for being unclear :)

WorksButNotTested: Ok thanks.

WorksButNotTested:
Sample code
```javascript
function make_blue (message) {
    var escape = String.fromCharCode(parseInt('033', 8));
    var prefix = "[0;34m";
    var suffix = "[0m\n";
    var blue = `${escape}${prefix}${message.trim()}${escape}${suffix}`
    return blue;
}

rpc.exports = {
    onFridaStdIn: function (message) {
        var stdOutData = make_blue (`> ${message.trim()}`)
        send(['frida:stdout', stdOutData]);
    }
}

send(['frida:stdout', make_blue("STARTED\n")]);

var info = {
    type: "log",
    level: "info",
    payload: "> error message"
}
_send (JSON.stringify(info), null);

var warning = {
    type: "log",
    level: "warning",
    payload: "> warning message"
}
_send (JSON.stringify(warning), null);

var error = {
    type: "log",
    level: "error",
    payload: "> error message"
}
_send (JSON.stringify(error), null);
```

inject/inject.vala
```diff
@@ -508,6 +573,56 @@ namespace Frida.Inject {
...
+		 * send(['frida:stderr', 'MESSAGE']);
+		 *
+		 * The resulting message will look as shown below. Note that we don't
+		 * use the parents object's `type` field since this is reserved for use
+		 * by the runtime itself.
+		 *
+		 * {"type":"send","payload":["frida:stdout","MESSAGE"]}
+		 */
+		private bool try_handle_stdout_message (Json.Object message) {
+			var payload = message.get_array_member ("payload");
```
oleavr, line 592: This API requires it to actually be an array if it exists. Might be easier to use a `JsonReader`.

inject/inject.vala
```diff
@@ -508,6 +573,56 @@ namespace Frida.Inject {
...
+			if (payload == null)
+				return false;
+
+			if (payload.get_length() != 2)
+				return false;
+
+			var type = payload.get_string_element (0);
```
oleavr, line 599: Like above, this requires the element to be a string.

inject/inject.vala
```diff
@@ -508,6 +573,56 @@ namespace Frida.Inject {
...
+		 * by the runtime itself.
+		 *
+		 * {"type":"send","payload":["frida:stdout","MESSAGE"]}
+		 */
+		private bool try_handle_stdout_message (Json.Object message) {
+			var payload = message.get_array_member ("payload");
+			if (payload == null)
+				return false;
+
+			if (payload.get_length() != 2)
```
oleavr, line 596:
```suggestion
			if (payload.get_length () != 2)
```

inject/inject.vala
```diff
@@ -508,6 +573,56 @@ namespace Frida.Inject {
...
+				return false;
+
+			var type = payload.get_string_element (0);
+			switch (type)
+			{
```
oleavr, line 601: Opening brace on wrong line.

inject/inject.vala
```diff
@@ -508,6 +573,56 @@ namespace Frida.Inject {
...
+				default:
+					return false;
+			}
+
+			var msg = payload.get_string_element (1);
+			if (msg == null)
+				return false;
+
+			switch (type)
+			{
```
oleavr, line 614: Opening brace on wrong line.

inject/inject.vala
```diff
@@ -225,6 +225,53 @@ namespace Frida.Inject {
...
+				}
+
+				string str_return = null;
+				try {
+					IOStatus status = inchan.read_line (out str_return, null, null);
+					if (status == IOStatus.EOF) {
+						loop.quit ();
+						return false;
+					}
+
```
oleavr, line 260:
```suggestion
```

inject/inject.vala
```diff
@@ -225,6 +225,53 @@ namespace Frida.Inject {
...
+			 * }
+			 */
+			var fd = stdin.fileno ();
+#if WINDOWS
+			var inchan = new IOChannel.win32_new_fd (fd);
+#else
+			var inchan = new IOChannel.unix_new (fd);
+#endif
+			inchan.add_watch (IOCondition.IN, (source, condition) => {
+				if (condition == IOCondition.HUP) {
```
oleavr, line 249: Braces can be omitted.

inject/inject.vala
```diff
@@ -225,6 +225,53 @@ namespace Frida.Inject {
...
+					return false;
+				}
+
+				string str_return = null;
+				try {
+					IOStatus status = inchan.read_line (out str_return, null, null);
+					if (status == IOStatus.EOF) {
```
oleavr, line 256: Let's move this outside the `try`.

oleavr (changes requested):

WorksButNotTested:
function make_blue (message) {
    var escape = String.fromCharCode(parseInt('033', 8));
    var prefix = "[0;34m";
    var suffix = "[0m\n";
    var blue = `${escape}${prefix}${message.trim()}${escape}${suffix}`
    return blue;
}

rpc.exports = {
    onFridaStdIn: function (message) {
        var stdOutData = make_blue (`> ${message.trim()}`)
        send(['frida:stdout', stdOutData]);
    }
}

send(['frida:stdout', make_blue("STARTED\n")]);

var info = {
    type: "log",
    level: "info",
    payload: "> error message"
}
_send (JSON.stringify(info), null);

var warning = {
    type: "log",
    level: "warning",
    payload: "> warning message"
}
_send (JSON.stringify(warning), null);

var error = {
    type: "log",
    level: "error",
    payload: "> error message"
}
_send (JSON.stringify(error), null);

send ("BLAH");
send (["test1", "test2", "test3"]);
send ([123, "test"]);
send (["frida:stdout", 123]);
send ([{ key: 123, value: "test"}, "abc"]);
send (["frida:stdout", { key: 123, value: "test"}]);
send ({ key: 123, value: "test"});
send (null);

WorksButNotTested:
The only input I can think which causes a crash now is when you call send without any parameters:
```javascript
send();
```

But then we don't even get as far as the `on_message` function.

oleavr:
Thanks! 🙌

PR #360, opened 2021-02-13
Fix frida-inject when process is not attached to a terminal

muhzii:


inject/inject.vala
```diff
@@ -245,6 +245,9 @@ namespace Frida.Inject {
 			var inchan = new IOChannel.unix_new (fd);
 #endif
 			inchan.add_watch (IOCondition.IN, (source, condition) => {
+				if (script_runner == null)
+					return false;
+
```
oleavr, line 250: This means we will stop processing `stdin` if input arrives before `start()` is done. We should instead move all of this logic to a method called at the end of `start ()`.

oleavr (changes requested):
Thanks!

inject/inject.vala
```diff
@@ -306,6 +280,34 @@ namespace Frida.Inject {
...
+#else
+			var inchan = new IOChannel.unix_new (fd);
+#endif
+			inchan.add_watch (IOCondition.IN, (source, condition) => {
+				if (script_runner.terminal_mode == COOKED)
+					return read_line (source, condition);
+				else
+					return read_raw (source, condition);
+			});
+		}
```
oleavr, line 309: This is lower level than `start ()`, so we should move it down – I'd say all the way down inside this class. The methods it relies on should also be moved down, below it. (Following [Newspaper Code Structure](https://kentcdodds.com/blog/newspaper-code-structure/)).

oleavr (changes requested):

oleavr (approved):

PR #363, opened 2021-02-25
frida-inject: Make the communication with script through stdin optional.

muhzii:
Having this enabled by default will prevent use cases such as running frida-inject as a background process.

Also, this currently conflicts with the --eternalize option where it is expected that the process will exit after loading the script.


oleavr:
Thanks!

PR #365, opened 2021-03-08
Support iOS app spawn on unc0ver 6.x

mrmacete:
By skipping the call to substituted which eventually results in resuming the app too early, only for the PIDs of the apps spawn by Frida.

Needs more testing to find out possible edge cases of this approach, but to my initial tests it fixes the spawn issue on unc0ver 6.x

### DIFF
--- a/src/darwin/agent/launchd.js
+++ b/src/darwin/agent/launchd.js
@@ -13,7 +13,8 @@ const reportCrashes = @REPORT_CRASHES@;
 let gating = false;
 const suspendedPids = new Set();
 
-let jbdPidsToIgnore = null;
+let pidsToIgnore = null;
+let _cachedInseterImpl = null;
 
 const substrateInvocations = new Set();
 const substratePidsPending = new Map();
@@ -106,8 +107,8 @@ Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dyli
 
     suspendedPids.add(pid);
 
-    if (jbdPidsToIgnore !== null)
-      jbdPidsToIgnore.add(pid);
+    if (pidsToIgnore !== null)
+      pidsToIgnore.add(pid);
 
     if (substrateInvocations.has(threadId)) {
       substratePidsPending.set(pid, notifyFridaBackend);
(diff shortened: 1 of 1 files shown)


src/darwin/agent/launchd.js
```diff
@@ -252,3 +271,44 @@ function findClosestMachHeader(address) {
...
   }
 }
+
+function findInserterResume() {
+  if (_cachedInseterImpl !== null) {
+    if (_cachedInseterImpl.isNull()) {
+      return null;
+    }
+    return _cachedInseterImpl;
+  }
```
oleavr, line 281: I guess there's only a single caller so there's no need to optimize this?

oleavr (changes requested):
w00t! 💥

src/darwin/agent/launchd.js
mrmacete: right!

src/darwin/agent/launchd.js
```diff
@@ -252,3 +270,34 @@ function findClosestMachHeader(address) {
     cur = cur.sub(4096);
   }
 }
+
+function findInserterResume() {
+  const candidates = Process.enumerateModules().filter(x => x.name === 'substitute-inserter.dylib');
+  if (candidates.length !== 1) {
+    return null;
+  }
```
oleavr, line 278:
```suggestion
  if (candidates.length !== 1)
    return null;
```

src/darwin/agent/launchd.js
```diff
@@ -252,3 +270,34 @@ function findClosestMachHeader(address) {
...
+
+  const { base, size } = candidates[0];
+  const signature = '8? 00 00 b4 e0 03 00 91 ?? ?? 00 9? e4 0f 40 b9 e0 03 00 91 e1 07 00 32 82 05 80 52 83 05 80 52 05 00 80 52';
+
+  const matches = Memory.scanSync(base, size, signature);
+  if (matches.length !== 1) {
+    return null;
+  }
```
oleavr, line 286:
```suggestion
  if (matches.length !== 1)
    return null;
```

src/darwin/agent/launchd.js
```diff
@@ -252,3 +270,34 @@ function findClosestMachHeader(address) {
...
+
+  let cursor = matches[0].address.sub(4);
+  const end = cursor.sub(1024);
+  while (cursor.compare(end) >= 0) {
+    try {
+      const instr = Instruction.parse(cursor);
+      if (instr.mnemonic.startsWith('ret')) {
+        return cursor.add(4).sign();
+      }
```
oleavr, line 295:
```suggestion
      if (instr.mnemonic.startsWith('ret'))
        return cursor.add(4).sign();
```

oleavr (changes requested):

oleavr (approved):
💥

PR #366, opened 2021-03-09
Workaround single-step delay

mrmacete:
It happens randomly that after hitting a single-step breakpoint, even if we disable it and continue, it still hits once.

(especially visible on iOS 14.x with unc0ver 6.x)

This change detects and reacts to this condition.

### DIFF
--- a/src/darwin/frida-helper-backend-glue.m
+++ b/src/darwin/frida-helper-backend-glue.m
@@ -160,6 +160,7 @@ struct _FridaSpawnInstance
   FridaBreakpoint breakpoints[FRIDA_MAX_BREAKPOINTS];
   FridaPagePoolEntry page_pool[FRIDA_MAX_PAGE_POOL];
   gint single_stepping;
+  GumAddress last_single_step_address;
   mach_vm_address_t lib_name;
   mach_vm_address_t bootstrapper_name;
   mach_vm_address_t fake_helpers;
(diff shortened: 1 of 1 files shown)


oleavr (approved):
👌🙌

oleavr:
Earlier today I ran into the same issue on the other edge – when enabling instead of disabling. So I cooked up the workaround in 7e023fff32fdc121698b100e512b00042b032852. It would then work most of the time, but still fail once in a while in some weird way. I was also experiencing system lockups on my M1 running macOS Monterey, which was strange.

Then when trying the new spawn gating on my old Intel MacBook, I ran into an issue that seemed kind of familiar: that `task_resume()` failed because it claimed the task port wasn't valid. And then I remembered the old observation that task ports shouldn't survive exec-transitions. So once I removed the task port caching from the spawn gating, it started working on the Intel system as well.

But still, I noticed weirdness when `resume()` was done without attaching in case of apps spawned (`xpcproxy` becoming the app). And then I realized there was the same caching issue there – cached task ports were being used from before the exec transition in that case.

Anyway, I quickly checked how much wall-clock time `task_for_pid()` typically takes on my Intel system, and it was pretty consistently less than a millisecond. So I went ahead and just removed all the caching logic. And then I realized that this single-step delay workaround that I added is no longer needed. And so far no system lockups (!).

So now I'm wondering if maybe also this issue has the same root cause, and should also be reverted?

PR #367, opened 2021-03-13
Fix build for new Xcode.

Manouchehri:
Doesn't seem to fix https://github.com/frida/frida/issues/1647 entirely, unless I'm still building it out of date by accident..

### DIFF
--- a/lib/payload/libc-shim.c
+++ b/lib/payload/libc-shim.c
@@ -456,7 +456,7 @@ _frida_set_errno (int n)
 # undef munmap
 
 ssize_t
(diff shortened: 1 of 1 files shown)


SolorzanoJose (approved):

lib/payload/libc-shim.c
```diff
@@ -456,7 +456,7 @@ _frida_set_errno (int n)
 # undef munmap
 
 ssize_t
-read (int fildes, const void * buf, size_t nbyte)
+read (int fildes, void * buf, size_t nbyte)
```
oleavr, line 459: Oops! Great catch.

PR #369, opened 2021-04-12
Shorten signature to match newer inserter too

mrmacete:


src/darwin/agent/launchd.js
```diff
@@ -277,7 +277,7 @@ function findInserterResume() {
     return null;
 
   const { base, size } = candidates[0];
-  const signature = '8? 00 00 b4 e0 03 00 91 ?? ?? 00 9? e4 0f 40 b9 e0 03 00 91 e1 07 00 32 82 05 80 52 83 05 80 52 05 00 80 52';
+  const signature = 'e0 03 00 91 e1 07 00 32 82 05 80 52 83 05 80 52 05 00 80 52';
```
oleavr, line 280: It seems like the first byte is different; how does that work on older versions? (Since we're not using a different offset from where this is found, I'm assuming this breaks support for older versions?)

oleavr (commented):
💥👏

src/darwin/agent/launchd.js
mrmacete: After finding the offset it walks back until the beginning of the function, by finding the RET instruction of the function which precedes this one.

So it doesn't really matter where this pattern is found, provided that's inside the interesting function.

oleavr: Ooh, sorry, I had the wrong context here. Let's get this landed! 💪

PR #370, opened 2021-04-14
Rework the function address resolution on Android

asabil:
Use the library base address retrieved using dladdr to find the path in /proc/:pid/maps instead of relying on the library path. This is to handle cases where the android environment is running inside a container.

Similarly, when resolving a linker function, use the already resolved linker base address to match the correct line in /proc/:pid/maps

### DIFF
--- a/src/linux/frida-helper-backend-glue.c
+++ b/src/linux/frida-helper-backend-glue.c
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
 }
 
 #ifdef HAVE_ANDROID
+static gboolean
+frida_find_library_path (pid_t pid, GumAddress base, gchar ** library_path)
+{
+  gboolean result = false;
+  gchar * maps_path;
+  FILE * fp;
+  const guint line_size = 1024 + PATH_MAX;
+  gchar * line, * path;
+
+  *library_path = NULL;
+
+  maps_path = g_strdup_printf ("/proc/%d/maps", pid);
+
+  fp = fopen (maps_path, "r");
+  g_assert (fp != NULL);
+
+  g_free (maps_path);
+
+  line = g_malloc (line_size);
+  path = g_malloc (PATH_MAX);
+
+  while (result == false && fgets (line, line_size, fp) != NULL)
+  {
+    GumAddress start;
+    gint n;
+
+    n = sscanf (line, "%" G_GINT64_MODIFIER "x-%*x %*s %*x %*s %*s %s", &start, path);
+    if (n == 1)
+      continue;
+    g_assert (n == 2);
+
+    if (path[0] == '[')
+      continue;
+
+    if (start == base)
+    {
+      result = true;
+      if (library_path != NULL)
+        *library_path = g_strdup (path);
+    }
+  }
+
+  g_free (path);
+  g_free (line);
+
+  fclose (fp);
+
+  return result;
+}
 
 static GumAddress
 frida_resolve_linker_address (pid_t pid, gpointer func)
 {
   Dl_info info;
-  const gchar * linker_path;
+  gchar * linker_path = NULL;
   GumAddress local_base, remote_base, remote_address;
 
-  if (dladdr (func, &info) != 0)
-    linker_path = info.dli_fname;
+  if (dladdr (func, &info) != 0) 
+    local_base = GUM_ADDRESS (info.dli_fbase);
   else
-    linker_path = gum_android_get_linker_module_details ()->path;
+    local_base = gum_android_get_linker_module_details ()->range->base_address;
 
-  local_base = frida_find_library_base (getpid (), linker_path, NULL);
   g_assert (local_base != 0);
+  if (!frida_find_library_path (getpid (), local_base, &linker_path))
+  {
+    g_critical ("Unable to locate library path; please file a bug");
+    g_abort ();
+  }
 
   remote_base = frida_find_library_base (pid, linker_path, NULL);
+
+  g_free (linker_path);
+
   if (remote_base == 0)
     return 0;
 
--- a/src/linux/frida-helper-backend-glue.c
+++ b/src/linux/frida-helper-backend-glue.c
@@ -3121,8 +3121,7 @@ frida_find_library_path (pid_t pid, GumAddress base, gchar ** library_path)
     if (start == base)
     {
       found = TRUE;
-      if (library_path != NULL)
-        *library_path = g_strdup (path);
+      *library_path = g_strdup (path);
     }
   }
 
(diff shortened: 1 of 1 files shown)


asabil:
The idea here is that we already know the base address of the function in our own process through the `dladdr()` function, we could use this base address to find the corresponding library path in `/proc/self/maps`, at this point knowing the library path, we use this library path when resolving the base address on the remote process.

Similarly, the base address of the linker has already been resolved, so we use that instead of using the linker path.

src/linux/frida-helper-backend-glue.c
```diff
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
 }
 
 #ifdef HAVE_ANDROID
```
oleavr, line 2845:
```suggestion
#ifdef HAVE_ANDROID

```

src/linux/frida-helper-backend-glue.c
```diff
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
...
+static gboolean
+frida_find_library_path (pid_t pid, GumAddress base, gchar ** library_path)
+{
+  gboolean result = false;
```
oleavr, line 2849:
```suggestion
  gboolean found = FALSE;
```

src/linux/frida-helper-backend-glue.c
```diff
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
...
+
+  fp = fopen (maps_path, "r");
+  g_assert (fp != NULL);
+
+  g_free (maps_path);
+
+  line = g_malloc (line_size);
+  path = g_malloc (PATH_MAX);
+
+  while (result == false && fgets (line, line_size, fp) != NULL)
```
oleavr, line 2867:
```suggestion
  while (!found && fgets (line, line_size, fp) != NULL)
```

src/linux/frida-helper-backend-glue.c
```diff
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
...
+    if (n == 1)
+      continue;
+    g_assert (n == 2);
+
+    if (path[0] == '[')
+      continue;
+
+    if (start == base)
+    {
+      result = true;
```
oleavr, line 2882:
```suggestion
      found = TRUE;
```

src/linux/frida-helper-backend-glue.c
```diff
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
...
+        *library_path = g_strdup (path);
+    }
+  }
+
+  g_free (path);
+  g_free (line);
+
+  fclose (fp);
+
+  return result;
```
oleavr, line 2893:
```suggestion
  return found;
```

src/linux/frida-helper-backend-glue.c
```diff
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
...
+}
```
oleavr, line 2894: Would be good to move this function after `frida_find_library_base()`, forward-declared with `G_GNUC_UNUSED`:

```C
G_GNUC_UNUSED static gboolean frida_find_library_path (pid_t pid, GumAddress base, gchar ** library_path);
```

src/linux/frida-helper-backend-glue.c
```diff
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
...
 
 static GumAddress
 frida_resolve_linker_address (pid_t pid, gpointer func)
 {
   Dl_info info;
-  const gchar * linker_path;
+  gchar * linker_path = NULL;
```
oleavr, line 2900:
```suggestion
  gchar * linker_path;
```

(As it's an out-argument and we trust the implementation to be well-behaved.)

src/linux/frida-helper-backend-glue.c
```diff
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
...
   GumAddress local_base, remote_base, remote_address;
 
-  if (dladdr (func, &info) != 0)
-    linker_path = info.dli_fname;
+  if (dladdr (func, &info) != 0) 
+    local_base = GUM_ADDRESS (info.dli_fbase);
   else
-    linker_path = gum_android_get_linker_module_details ()->path;
+    local_base = gum_android_get_linker_module_details ()->range->base_address;
```
oleavr, line 2907:
```suggestion
```

src/linux/frida-helper-backend-glue.c
```diff
@@ -2843,23 +2843,79 @@ frida_find_libc_base (pid_t pid)
...
 
-  local_base = frida_find_library_base (getpid (), linker_path, NULL);
   g_assert (local_base != 0);
```
oleavr, line 2908:
```suggestion
  g_assert (local_base != 0);

```

oleavr (changes requested):
w00t! Thanks! 💥

src/linux/frida-helper-backend-glue.c
asabil: On my side this resulted in a compiler warning

src/linux/frida-helper-backend-glue.c
asabil: So you don't want it to be guarded by an `#ifdef HAVE_ANDROID` ?

src/linux/frida-helper-backend-glue.c
oleavr: Oh! So it detects that `frida_find_library_path()` may not always initialize the out-argument?

src/linux/frida-helper-backend-glue.c
```diff
@@ -2844,22 +2844,79 @@ frida_find_libc_base (pid_t pid)
...
+      continue;
+    g_assert (n == 2);
+
+    if (path[0] == '[')
+      continue;
+
+    if (start == base)
+    {
+      found = TRUE;
+      if (library_path != NULL)
```
oleavr, line 2884: This check can be removed as it's not a nullable argument.

src/linux/frida-helper-backend-glue.c
oleavr: Indeed, it feels like it belongs next to `frida_find_library_base()`. Just need a `G_GNUC_UNUSED` declaration to avoid a compiler warning on non-Android. (The linker will eliminate the unused code anyway.)

src/linux/frida-helper-backend-glue.c
asabil: `frida_find_library_path ()` always initialises to NULL

asabil: I just tried again and I don't get the warning, I am sure I saw it earlier... Maybe it was in the other work in progress branch where I used a slightly different strategy.

oleavr (approved):
💪

PR #373, opened 2021-05-07
Expose system_parameters via HostSession API

hexploitable:
This MR, based on `feature/portal` exposes a system parameters hashmap via `HostSession`.

For now, the only key-value pair within is the platform, determined at compile time since we already know it. This object can be extended going forward if we think of anything extra we'd like to include.

I'll separately push the PR's for `frida-python` and `frida-node` but here's the anticipated usage:
```python
>>> import frida
>>> d = frida.enumerate_devices()
>>> d
[Device(id="local", name="Local System", type='local'), Device(id="socket", name="Local Socket", type='remote'), Device(id="REDACTED", name="iPhone", type='usb'), Device(id="REDACTED", name="Pixel 3", type='usb')]
>>> a = d[3]
>>> a
Device(id="REDACTED", name="Pixel 3", type='usb')
>>> i = d[2]
>>> m = d[0]
>>> m.query_system_parameters()
{'platform': 'macos'}
>>> i.query_system_parameters()
{'platform': 'ios'}
>>> a.query_system_parameters()
{'platform': 'android'}
>>>
```


lib/base/session.vala
```diff
@@ -20,6 +20,7 @@ namespace Frida {
 			Cancellable? cancellable) throws GLib.Error;
 		public abstract async InjectorPayloadId inject_library_blob (uint pid, uint8[] blob, string entrypoint, string data,
 			Cancellable? cancellable) throws GLib.Error;
+		public abstract async HashTable<string, Variant>? query_system_parameters (Cancellable? cancellable) throws GLib.Error;
```
oleavr, line 23:
```suggestion
		public abstract async HashTable<string, Variant> query_system_parameters (Cancellable? cancellable) throws GLib.Error;
```

Would also be good to move it right before `get_frontmost_application ()`.

lib/gadget/gadget.vala
```diff
@@ -1879,6 +1879,15 @@ namespace Frida.Gadget {
 				if (pid != this_process.pid)
 					throw new Error.NOT_SUPPORTED ("Unable to act on other processes when embedded");
 			}
+
+			public async HashTable<string, Variant>? query_system_parameters (Cancellable? cancellable) throws Error, IOError {
+				HostSession host_session = this;
```
oleavr, line 1884: This will result in an infinite loop as it's querying itself.

src/control-service.vala
```diff
@@ -638,6 +638,14 @@ namespace Frida {
 					Cancellable? cancellable) throws GLib.Error {
 				return yield parent.host_session.inject_library_blob (pid, blob, entrypoint, data, cancellable);
 			}
+
+			public async HashTable<string, Variant>? query_system_parameters (Cancellable? cancellable) throws Error, IOError {
+				try {
+					return yield parent.host_session.query_system_parameters (cancellable);
+				} catch (GLib.Error e) {
+					return null;
```
oleavr, line 646: Shouldn't catch the exception here, but instead declare that this method throws `GLib.Error` to match the method we're calling.

src/frida.vala
```diff
@@ -1379,6 +1385,20 @@ namespace Frida {
...
+				return yield host_session.query_system_parameters (cancellable);
+			} catch (GLib.Error e) {
+				return null;
+			}
+		}
+
+		public HashTable<string, Variant>? query_system_parameters_sync (Cancellable? cancellable = null) throws Error, IOError {
+			var task = create<QuerySystemParametersTask> ();
+			return task.execute (cancellable);
+		}
```
oleavr, line 1401: These should go just before `get_frontmost_application ()`.

src/frida.vala
```diff
@@ -1143,6 +1143,12 @@ namespace Frida {
 			}
 		}
 
+		private class QuerySystemParametersTask : DeviceTask<HashTable<string, Variant>?> {
+			protected override async HashTable<string, Variant>? perform_operation () throws Error, IOError {
+				return yield parent.query_system_parameters (cancellable);
+			}
+		}
```
oleavr, line 1150: This should be moved to the two methods it belongs to, following the order we're using for other methods.

src/host-session-service.vala
```diff
@@ -345,6 +345,29 @@ namespace Frida {
...
+			#elif LINUX
+				Variant? platform = new Variant ("s", "linux");
+			#elif WINDOWS
+				Variant? platform = new Variant ("s", "windows");
+			#else
+				Variant? platform = new Variant ("s", "unknown");
+			#endif
+			
+			var system_parameters = new HashTable<string, Variant> (str_hash, str_equal);
+			system_parameters.insert ("platform", platform);
```
oleavr, line 366:
```suggestion
			system_parameters["os"] = os;
```

To make it more specific than `Process.platform` and avoid confusion that it's named the same but has different granularity.

src/host-session-service.vala
```diff
@@ -345,6 +345,29 @@ namespace Frida {
 			throw new Error.INVALID_OPERATION ("Only meant to be implemented by services");
 		}
 
+		public async HashTable<string, Variant>? query_system_parameters (Cancellable? cancellable) throws Error, IOError {
+			#if DARWIN && !IOS
+				Variant? platform = new Variant ("s", "macos");
+			#elif IOS
+				Variant? platform = new Variant ("s", "ios");
+			#elif ANDROID && (X86 || X86_64)
+				Variant? platform = new Variant ("s", "android_emulated");
```
oleavr, line 354: This is wrong in some cases, like with Android x86 tablets, so we shouldn't try to disambiguate here.

src/host-session-service.vala
```diff
@@ -345,6 +345,29 @@ namespace Frida {
 			throw new Error.INVALID_OPERATION ("Only meant to be implemented by services");
 		}
 
+		public async HashTable<string, Variant>? query_system_parameters (Cancellable? cancellable) throws Error, IOError {
+			#if DARWIN && !IOS
+				Variant? platform = new Variant ("s", "macos");
```
oleavr, line 350: Would be good to declare `string os;` at the top, and then just assign a different value to it.

src/droidy/droidy-host-session.vala
```diff
@@ -210,6 +210,15 @@ namespace Frida {
 			);
 		}
 
+		public async HashTable<string, Variant>? query_system_parameters (Cancellable? cancellable) throws Error, IOError {
+			try {
+				var server = yield get_remote_server (cancellable);
+				return yield server.session.query_system_parameters (cancellable);
```
oleavr, line 216: We will also need a fallback here for non-rooted Android (and similarly in the Fruity backend for jailed iOS).

oleavr (changes requested):
w00t! 🤘

PR #374, opened 2021-05-07
Fix ___error symbol name in fruity injector

mrmacete:
Noticed while debugging other issues that `get_errno_storage` was always NULL, fixed by adding the missing underscore to the mangled symbol name.


oleavr:
Thanks! 🤘

PR #375, opened 2021-05-26
Remove bashisms from iOS maintainer scripts

nyuszika7h:
Procursus-based jailbreaks such as Odyssey, Taurine and odysseyra1n use dash as `/bin/sh`.
The use of bashisms without a `#!/bin/bash` shebang causes it to fail to start/stop frida-server.


oleavr:
Yay, nice work! 🙌 Thanks a lot! ❤️

PR #388, opened 2021-08-19
Fix agent call fail when FRIDA_V8=disabled

rev1si0n:
when build android server with FRIDA_V8=disabled, frida-python will not work
self._impl.enumerate_processes(*args, **kwargs): frida.NotSupportedError: v8 runtime not available due to build configuration.
this is not a graceful fix.


oleavr:
Thanks for taking a stab at this! In order to land this, we will have to check whether V8 is available (meson.build) and specify the runtime based on that. Otherwise this will result in a performance regression on Android.

PR #390, opened 2021-09-01
Update pipe-darwin.c to use correct error string

mattshockl:
Update pipe-darwin.c to use bootstrap_strerror instead of mach_error_string for errors coming from bootstrap_look_up.  Currently the error printed every time is "Unknown error code"


lib/pipe/pipe-darwin.c
```diff
@@ -31,6 +37,7 @@ typedef char frida_pipe_uuid_t[36 + 1];
 #include "piped-client.c"
 
 extern kern_return_t bootstrap_look_up (mach_port_t bootstrap_port, const char * service_name, mach_port_t * service_port);
+extern const char *bootstrap_strerror (kern_return_t r);
```
oleavr, line 40:
```suggestion
extern const char * bootstrap_strerror (kern_return_t kr);
```

oleavr (approved):
Thanks, great catch! 🙌

PR #391, opened 2021-09-02
Add 'no-bootstrap-service' option to frida-server

mattshockl:
Add option to not register a bootstrap service and just rely on the stashed mach port. For some unknown reason to me, [the sandbox extension code](https://github.com/frida/frida-core/blob/master/src/darwin/frida-helper-backend-glue.m) in the `frida_darwin_helper_backend_make_pipe_endpoints` function fails transparently for some processes (apsd, imagent, etc) on Big Sur. With no fallback options, it makes it impossible to hook these applications. This option allows a user to specify to not use the bootstrap server and instead fall back on just the traditional mach port IPC.

This also seems to be causing other people issues:
https://github.com/frida/frida/issues/1829
https://github.com/frida/frida/issues/1830
https://github.com/frida/frida/issues/1719

Here's an example with apsd without `--no-bootstrap-service`
Screen Shot 2021-09-02 at 5 14 42 PM

Here's an example with apsd with `--no-bootstrap-service`
Screen Shot 2021-09-02 at 5 18 02 PM

### DIFF
--- a/lib/pipe/pipe-darwin.c
+++ b/lib/pipe/pipe-darwin.c
@@ -31,6 +31,7 @@ typedef char frida_pipe_uuid_t[36 + 1];
 #include "piped-client.c"
 
 extern kern_return_t bootstrap_look_up (mach_port_t bootstrap_port, const char * service_name, mach_port_t * service_port);
+extern const char *bootstrap_strerror (kern_return_t r);
 extern int fileport_makeport (int fd, mach_port_t * port);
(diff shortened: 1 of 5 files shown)


oleavr:
Thanks! 🎉  I think we will need to find a solution that also works on Monterey / iOS 15, where guarded Mach ports prevent us from using our traditional mach port IPC. So given the (hopefully) temporary nature of this workaround, I think we should use an environment variable instead, as this will also work for non-frida-server use-cases.

PR #392, opened 2021-09-02
Add NO_BOOTSTRAP_SERVICE environment variable functionality

mattshockl:
Based on feedback from https://github.com/frida/frida-core/pull/391, changed the functionality to just look for a NO_BOOTSTRAP_SERVICE environment variable as a temporary solution to this issue.

### DIFF
--- a/lib/pipe/pipe-darwin.c
+++ b/lib/pipe/pipe-darwin.c
@@ -31,6 +31,7 @@ typedef char frida_pipe_uuid_t[36 + 1];
 #include "piped-client.c"
 
 extern kern_return_t bootstrap_look_up (mach_port_t bootstrap_port, const char * service_name, mach_port_t * service_port);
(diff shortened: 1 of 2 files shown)


oleavr:
Thanks! I couldn't stop thinking about how to solve this properly, and then it eventually struck me that maybe this new IPC scheme isn't actually needed. It was introduced to support iOS 15, but only after fixing the first stumbling block, during which I may have gained an incorrect understanding of guarded Mach ports, in turn resulting in me thinking I also had to rework the IPC part. I've reverted this code for now, so there's no longer any bootstrap service. I'm still waiting for feedback from someone with access to jailbroken iOS 15 and just started releasing Frida 15.1.0 with these changes. Let's see if we need to roll it back and revisit this. But for now let's close this and hope for the best 🤞

PR #393, opened 2021-09-03
Add interaction.parameters to gadget in connect mode

mrmacete:
These parameters are then "reflected" into app's info under `parameters.parameters`

### DIFF
--- a/lib/gadget/gadget.vala
+++ b/lib/gadget/gadget.vala
@@ -2007,7 +2007,7 @@ namespace Frida.Gadget {
 			var interaction = config.interaction as ConnectInteraction;
 
 			try {
-				info.parameters["parameters"] = Json.gvariant_deserialize (interaction.parameters, null);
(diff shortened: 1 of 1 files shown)


lib/gadget/gadget.vala
```diff
@@ -1968,6 +2002,19 @@ namespace Frida.Gadget {
 			client.kill.connect (Frida.Gadget.kill);
 		}
 
+		protected override HostApplicationInfo compute_app_info () {
+			var info = base.compute_app_info ();
+			var interaction = config.interaction as ConnectInteraction;
+
+			try {
+				info.parameters["parameters"] = Json.gvariant_deserialize (interaction.parameters, null);
```
oleavr, line 2010:
```suggestion
				info.parameters["config"] = Json.gvariant_deserialize (interaction.parameters, null);
```

Another option is `"user"`.

oleavr (changes requested):
Thanks! 🤘

PR #394, opened 2021-09-04
Improve frida-inject to start specific activity

hunterli:
To resolve issue: Unable to find a front-door activity

### DIFF
--- a/inject/inject.vala
+++ b/inject/inject.vala
@@ -3,6 +3,7 @@ namespace Frida.Inject {
 
 	private static string? device_id;
 	private static string? spawn_file;
+	private static string? activity_name;
 	private static int target_pid = -1;
 	private static string? target_name;
 	private static string? realm_str;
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! I think it would be better to add a CLI option to add aux options generically. This way we won't have to add CLI options for all the other OS-specific options. It would be great to do what frida-tools [does](https://github.com/frida/frida-tools/blob/16664d6b629713eeb7b82653d8fddc5b7702b707/frida_tools/application.py#L144-L145).

hunterli:
> Thanks! I think it would be better to add a CLI option to add aux options generically. This way we won't have to add CLI options for all the other OS-specific options. It would be great to do what frida-tools [does](https://github.com/frida/frida-tools/blob/16664d6b629713eeb7b82653d8fddc5b7702b707/frida_tools/application.py#L144-L145).

OK

PR #396, opened 2021-09-22
control-service: fix signal wiring

mrmacete:
Route `output`, `child_added`, `child_removed`, `process_crashed` and `uninjected` signals to all control channel peers.

### DIFF
--- a/src/control-service.vala
+++ b/src/control-service.vala
@@ -254,9 +254,7 @@ namespace Frida {
 		}
 
 		private Gee.Iterator<ControlChannel> all_control_channels () {
-			return (Gee.Iterator<ControlChannel>) peers.values.filter (peer => {
-				return (peer as ControlChannel) != null;
-			});
+			return (Gee.Iterator<ControlChannel>) peers.values.filter (peer => peer is ControlChannel);
 		}
 
 		private async void enable_spawn_gating (ControlChannel requester) throws GLib.Error {
--- a/src/control-service.vala
+++ b/src/control-service.vala
@@ -78,11 +78,12 @@ namespace Frida {
 
 		construct {
 			host_session.spawn_added.connect (notify_spawn_added);
-			host_session.agent_session_detached.connect (on_agent_session_detached);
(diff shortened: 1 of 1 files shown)


src/control-service.vala
```diff
@@ -249,6 +253,12 @@ namespace Frida {
 			}
 		}
 
+		private Gee.Iterator<ControlChannel> all_control_channels () {
+			return (Gee.Iterator<ControlChannel>) peers.values.filter (peer => {
+				return (peer as ControlChannel) != null;
+			});
```
oleavr, line 259: Should be possible to simplify this to:

```suggestion
			return (Gee.Iterator<ControlChannel>) peers.values.filter (peer => peer is ControlChannel);
```

src/control-service.vala
```diff
@@ -79,6 +79,10 @@ namespace Frida {
 		construct {
 			host_session.spawn_added.connect (notify_spawn_added);
 			host_session.agent_session_detached.connect (on_agent_session_detached);
+			host_session.child_added.connect (notify_child_added);
+			host_session.child_removed.connect (notify_child_removed);
+			host_session.process_crashed.connect (notify_process_crashed);
+			host_session.output.connect (notify_output);
```
oleavr, line 85: - Would be good to add these between `spawn_added` and `agent_session_detached`, so they follow the same order as in the `HostSession` interface.
- While at it we should also wire up `uninjected`.

src/control-service.vala
```diff
@@ -383,6 +393,34 @@ namespace Frida {
 				channel.spawn_removed (info);
 		}
 
+		private void notify_output (uint pid, int fd, uint8[] data) {
+			all_control_channels ().foreach (channel => {
+				channel.output (pid, fd, data);
+				return true;
+			});
+		}
```
oleavr, line 401: These should ideally be in the same order as the signal definitions in the `HostSession` interface.

oleavr (changes requested):
Thanks! 🤘

PR #401, opened 2021-11-01
portal-service: Check NULL session entry in teardown_cluster_node()

mrmacete:
Apparently `sessions[id]` can be NULL under certain conditions, the check avoids crashing in that case.


mrmacete:
Closing in favour of https://github.com/frida/frida-core/pull/402

PR #402, opened 2021-11-02
portal-service: Always remove ClusterNode and ControlChannel sessions

mrmacete:
Whenever a session id is unset from PortalService. This avoids both NULL derefereces and leaks.

### DIFF
--- a/src/portal-service.vala
+++ b/src/portal-service.vala
@@ -489,12 +489,13 @@ namespace Frida {
 				AgentSession? session = entry.session;
 				if (entry.persist_timeout == 0 || session == null) {
 					sessions.unset (id);
-					if (session != null)
-						session.close.begin (io_cancellable);
 
 					ClusterNode? node = entry.node;
 					if (node != null)
 						node.sessions.remove (id);
+
+					if (session != null)
+						session.close.begin (io_cancellable);
 				} else {
 					entry.detach_controller ();
 					session.interrupt.begin (io_cancellable);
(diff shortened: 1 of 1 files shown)


src/portal-service.vala
```diff
@@ -837,6 +841,14 @@ namespace Frida {
...
 		private void on_agent_session_expired (AgentSessionEntry entry) {
 			sessions.unset (entry.id);
+
+			ClusterNode? node = entry.node;
+			if (node != null)
+				node.sessions.remove (entry.id);
+
+			ControlChannel? controller = entry.controller;
+			if (controller != null)
+				controller.sessions.remove (entry.id);
```
mrmacete, line 851: Not super sure this is needed, though.

oleavr: Ah yes indeed, since the timer gets started by `detach_node_and_controller ()` and `detach_controller ()`, and in the former case `teardown_cluster_node ()` does this, and in the latter case we are tearing down the `ControlChannel`.

src/portal-service.vala
```diff
@@ -491,6 +491,10 @@ namespace Frida {
 					sessions.unset (id);
 					if (session != null)
 						session.close.begin (io_cancellable);
+
+					ClusterNode? node = entry.node;
+					if (node != null)
+						node.sessions.remove (id);
```
oleavr, line 497: Let's move these between `sessions.unset ()` and the `close.begin ()`.

oleavr (changes requested):
Since it took me some time to get to this, I'll take care of these tweaks now and land this:

PR #406, opened 2022-01-10
droidy: Add AXML decoder

meme:


src/droidy/axml.vala
```diff
@@ -0,0 +1,336 @@
...
+				b.append ("\t");
+			b.append_printf ("</%s>", name);
+			if (depth != 0)
+				b.append ("\n");
+
+			return b.str;
+		}
+	}
+
+	private class EndElement : Object {
```
oleavr, line 124: This isn't used currently, should it be kept for later?

oleavr (commented):
Thanks, nice work! Sorry it took me so long to get to this.

PR #422, opened 2022-07-20
Fix use after free in crash delivery

muhzii:
In case a crash is reported after a Java crash has already been delivered.
E.g. in apps that set their own uncaught exception handler, rendering
uncaught exceptions non-fatal.


oleavr:
Thanks! :fire:

PR #426, opened 2022-07-26
Handle dashes in hostnames

0x3c3e:
Domain like `ab-c.com` will be discarded by previous regex, as it contains a dash.


oleavr:
Thanks! Landed with some minor tweaks as d370ff41f57e278ed3ed96dacaaf9a5d54b90e5b and 3f1ec9985a16d0d9833aaec8df2de2f01020927c.

Issue #429, opened 2022-08-18
daemonized frida server parent doesn't exit for some failures


pachoo:
I noticed that the parent frida-server process does not always exit when daemonized.  

Example
Frida 15.2.2
Device: Pixel 4a, Android 10, rooted

adb shell 1:
```
sunfish:/ # /data/local/tmp/frida-server-15.2.2-android-arm64 -D                                                                                                                   
sunfish:/ # ps -Af | grep frida
root         31218     1 6 12:46:18 ?     00:00:00 frida-server-15.2.2-android-arm64 -D
root         31234     1 1 12:46:19 ?     00:00:00 frida-helper-32 unix:abstract=/data/local/tmp/re.frida.server/2ffc05b6-e447-46bc-9d7d-0eaeec39c8c8
# Second start error, parent does not exit
sunfish:/ # /data/local/tmp/frida-server-15.2.2-android-arm64 -D                                                                                                                   
Unable to start: Could not listen on address 127.0.0.1, port 27042: Error binding to address 127.0.0.1:27042: Address already in use
```

adb shell 2
```
sunfish:/ # ps -Af | grep frida                                                                                                                                                  
root         31218     1 1 12:46:17 ?     00:00:00 frida-server-15.2.2-android-arm64 -D
root         31234     1 0 12:46:18 ?     00:00:00 frida-helper-32 unix:abstract=/data/local/tmp/re.frida.server/2ffc05b6-e447-46bc-9d7d-0eaeec39c8c8
# parent running
root         31264 30723 0 12:46:29 /sbin/.magisk/pts/2 00:00:00 frida-server-15.2.2-android-arm64 -D
root         31265 31264 0 12:46:29 136:2 00:00:00 [frida-server-15]
```

Taking a quick look at the source it looks like the child/daemon may not send the failure status back to the parent which is waiting to read it.

pachoo:
some late cleaning here, fixed in the above PR.  closing.

PR #430, opened 2022-08-18
frida-server: Fix daemon parent ready() fail exit

pachoo:
The frida-server Application class can now signal success or failure on `ready()` to allow for the daemon's parent to exit on failures.
See:  https://github.com/frida/frida-core/issues/429

Example
```
sunfish:/ # /data/local/tmp/frida-server -D
# This failure now triggers an exit
sunfish:/ # /data/local/tmp/frida-server -D                                                                                                                                        
Unable to start: Could not listen on address 127.0.0.1, port 27042: Error binding to address 127.0.0.1:27042: Address already in use
1|sunfish:/ # 
```

### DIFF
--- a/portal/portal.vala
+++ b/portal/portal.vala
@@ -133,8 +133,8 @@ namespace Frida.Portal {
 		});
 
 		if (on_ready != null) {
-			application.ready.connect (() => {
-				on_ready (true);
+			application.ready.connect (success => {
+				on_ready (success);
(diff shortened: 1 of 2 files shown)


oleavr:
Thank you, this is awesome! (And sorry for the delay here, took me a while to escape the rabbit hole that I got stuck in :sweat_smile:)

PR #434, opened 2022-09-08
Remove usbmuxd confusing envvars

as0ler:
This MR is based on https://github.com/frida/frida-core/pull/426.

It removes USBMUXD_SERVER_ADDRESS and USBMUXD_SERVER_PORT environment variables because can be confusing with USBMUXD_SOCKET_ADDRESS which is the one used by `libusbmuxd` too.


oleavr:
Thanks! Landed as 57f9a4e59e1473af324f03c78321f2ae44bed397.

PR #439, opened 2022-10-22
Add missing openssl dependency

yotamN:
I'm not sure how it worked for others. Maybe when `gioopenssl` is present it also solve the required openssl headers?


oleavr:
Thanks!

PR #441, opened 2022-10-28
cat /proc/22001/maps | grep http://libc.so

lx866:
6e3e1cd000-6e3e2c7000 r--s 00000000 07:08 38                             /apex/com.android.runtime/lib64/bionic/libc.so
71eb6c2000-71eb6ff000 r--p 00000000 07:08 38                             /apex/com.android.runtime/lib64/bionic/libc.so
71eb6ff000-71eb781000 r-xp 0003d000 07:08 38                             /apex/com.android.runtime/lib64/bionic/libc.so
71eb781000-71eb786000 r--p 000bf000 07:08 38                             /apex/com.android.runtime/lib64/bionic/libc.so
71eb786000-71eb787000 rw-p 000c3000 07:08 38                             /apex/com.android.runtime/lib64/bionic/libc.so

the first line mapping of the http://libc.so it's shared , we can't start from it we need to start from the second line


oleavr:
Thanks! 🙌

PR #446, opened 2022-12-06
Added support to detect when /tmp is mounted noexec

WorksButNotTested:


src/linux/system-linux.c
```diff
@@ -151,12 +158,156 @@ frida_system_kill (guint pid)
...
+  for (
+    current = getmntent_r(fp, &mntent, mntent_buf, sizeof (mntent_buf));
+    current != NULL;
+    current = getmntent_r(fp, &mntent, mntent_buf, sizeof (mntent_buf)))
+  {
+    /*
+     * In Linux 2.6 rootfs mount always exists, and it makes sense to always
+     * ignore it.
+     */
+    if (g_strcmp0(current->mnt_fsname, "rootfs") == 0)
```
oleavr, line 236: (Here and elsewhere.) Can `current->mnt_fsname` be `NULL`? If not, we should be using `strcmp()`.

src/linux/system-linux.c
```diff
@@ -151,12 +158,156 @@ frida_system_kill (guint pid)
...
+frida_is_mount_noexec (struct mntent * mntent)
+{
+  if (hasmntopt (mntent, "noexec") == NULL)
+    return FALSE;
+  else
+    return TRUE;
+}
+
+static gchar *
+get_application_directory (void)
```
oleavr, line 295: This PR has some style deviations. Basically:
- Functions should be prefixed with `frida_` even if internal.
- Missing spaces around `*`.
- Space before `(`.
- Redundant blocks when if/else only has a single statement.

src/linux/system-linux.c
```diff
@@ -151,12 +158,156 @@ frida_system_kill (guint pid)
...
+
+  exe_path = g_file_read_link ("/proc/self/exe", NULL);
+  if (exe_path == NULL)
+    goto beach;
+
+  exe_dir = g_path_get_dirname (exe_path);
+
+beach:
+  if (exe_path != NULL)
+    g_free (exe_path);
```
oleavr, line 308: No need for the `if` here, as `g_free()` is a no-op in case of `NULL`.

oleavr (changes requested):
Thanks! 🙌

src/linux/system-linux.c
WorksButNotTested: The `man` page for `getmntent_r` doesn't explicitly say that the structure member can be NULL, or that it can't. Given we may find ourselves running on many different versions of `libc` and perhaps different variants too (e.g. uClibc, musl) is it prudent to be safe?

oleavr: Just checked the sources of glibc and uClibc, including an ancient glibc version, and determined that the worst-case is an empty string. So it seems safe to assume these are never `NULL`. I'll change these to `strcmp()` to make the code a bit clearer.

oleavr:
Just landed as aa9516715663c1cbcbd91235e62f30c89d6fbdd1, simplified to use GLib and Gum for the heavy lifting. Thanks! :raised_hands:

PR #456, opened 2023-02-16
Fix iOS15 spawn issue

as0ler:
- Resolves _gProcessInfo mangled symbol when the original one is not found.
- Fix libSystem initialization issue in early instrumentation for iOS 15.


mrmacete:
Here are some more details about the libSystem initialisation issue.

It happens when we want to place a breakpoint to "step out" of libSystem's `libSystem_initializer()`. In order to do that we get the return address from the stack and place a breakpoint there.

On some versions of iOS 15 (example 15.5) this works because at that point we're stopped inside a tail call (which is just a direct branch):

Screenshot 2023-02-16 at 10 33 41

On other iOS 15 versions (like iOS 15.1), instead this isn't a tail call but a proper function call, which implies one more return address on the stack:

Screenshot 2023-02-16 at 10 33 56

so in that case we have to "dig" one step deeper in the stack to get to the first return address which is outside libSystem initializer and a safe place where we know libSystem is definitely inited.

mrmacete:
so apparently the `SYMBOL_RESOLVER_CODE` blob changed even if i didn't touch the corresponding helper code, i guess it's for different compiler behaviour but i'm not super sure.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1843,6 +1845,16 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
 
   dyld = gum_darwin_module_new_from_memory ("/usr/lib/dyld", task, dyld_header, GUM_DARWIN_MODULE_FLAGS_NONE, NULL);
   instance->dyld = dyld;
+  instance->dyld_size = 0;
+  i = 0;
+  while ((segment = gum_darwin_module_get_nth_segment (dyld, i++)) != NULL)
+  {
+    if (strncmp (segment->name, "__TEXT", sizeof (segment->name)))
```
oleavr, line 1852:
```suggestion
    if (strcmp (segment->name, "__TEXT") == 0)
```

This would work too right? And I guess this logic was inverted since it was checking for non-zero?

src/darwin/frida-helper-backend-glue.m
```diff
@@ -2927,21 +2943,37 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
       GumAddress frame_pointer = 0;
       guint64 * ret_addr_data;
       GumAddress libsystem_initializer_caller;
+      gboolean falls_into_dyld;
```
oleavr, line 2946:
```suggestion
      gboolean falls_within_dyld;
```

(At least I think that's the English idiom.)

src/fruity/injector.vala
```diff
@@ -838,8 +838,9 @@ namespace Frida.Fruity.Injector {
 		private async void ensure_libsystem_initialized_for_dyld_v4_and_above (uint64 libdyld_initialize,
 				Cancellable? cancellable) throws GLib.Error {
 			uint64? process_info_ptr = dyld_symbols["_gProcessInfo"];
-			if (process_info_ptr == null)
-				throw new Error.UNSUPPORTED ("Missing gProcessInfo");
+			if (process_info_ptr == null) {
+				process_info_ptr = dyld_symbols["__ZN5dyld412gProcessInfoE"];
+			}
```
oleavr, line 843: Redundant block.

src/fruity/injector.vala
```diff
@@ -882,10 +885,20 @@ namespace Frida.Fruity.Injector {
 			assert (hit_breakpoint == init_breakpoint);
 			yield init_breakpoint.remove (cancellable);
 
+			uint64 dyld_size = dyld_symbols["dyld_size"];
 			uint64 frame_here = exception.context["fp"];
 			uint64 frame_above = yield lldb.read_pointer (frame_here, cancellable);
-			uint64 libsystem_initializer_caller = yield lldb.read_pointer (frame_above + 8, cancellable);
-			libsystem_initializer_caller = lldb.strip_code_address (libsystem_initializer_caller);
+			uint64 libsystem_initializer_caller = 0;
+			bool falls_into_dyld = false;
```
oleavr, line 892:
```suggestion
			bool falls_within_dyld = false;
```

oleavr (changes requested):
> so apparently the SYMBOL_RESOLVER_CODE blob changed even if i didn't touch the corresponding helper code, i guess it's for different compiler behaviour but i'm not super sure.

Yeah just using a slightly different version of Xcode is likely to do that. I think it's better to omit those changes when we know we didn't touch any code affecting a given helper.

src/darwin/frida-helper-backend-glue.m
mrmacete: wow yeah 🤔

mrmacete:
should be fine, also reverted the unexpected changes to `SYMBOL_RESOLVER_CODE`

PR #464, opened 2023-04-17
Fix ProcMapsEntry.find_by_path return null

chouex:
The regex used in ProcMapsEntry.find_by_path will not match libc.so line when its major number grater then 0xff (eg: 103:04), This will cause ProcMapsEntry.find_by_path to return null, and frida-server fail to start.

reference:
https://www.kernel.org/doc/Documentation/admin-guide/devices.txt

fixed https://github.com/frida/frida/issues/2484

### DIFF
--- a/src/linux/frida-helper-backend.vala
+++ b/src/linux/frida-helper-backend.vala
@@ -3288,7 +3288,7 @@ namespace Frida {
 					return;
 				}
 
-				if (!/^([0-9a-f]+)-([0-9a-f]+) \S{4} [0-9a-f]+ ([0-9a-f]{2}:[0-9a-f]{2} \d+) +([^\n]+)$/m.match (contents,
+				if (!/^([0-9a-f]+)-([0-9a-f]+) \S{4} [0-9a-f]+ ([0-9a-f]{2,}:[0-9a-f]{2} \d+) +([^\n]+)$/m.match (contents,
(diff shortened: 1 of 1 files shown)


chouex:
Maps file sample
image
After fix 
image

src/linux/frida-helper-backend.vala
```diff
@@ -3288,7 +3288,7 @@ namespace Frida {
 					return;
 				}
 
-				if (!/^([0-9a-f]+)-([0-9a-f]+) \S{4} [0-9a-f]+ ([0-9a-f]{2}:[0-9a-f]{2} \d+) +([^\n]+)$/m.match (contents,
+				if (!/^([0-9a-f]+)-([0-9a-f]+) \S{4} [0-9a-f]+ ([0-9a-f]{2,}:[0-9a-f]{2} \d+) +([^\n]+)$/m.match (contents,
```
oleavr, line 3291:
```suggestion
				if (!/^([0-9a-f]+)-([0-9a-f]+) \S{4} [0-9a-f]+ ([0-9a-f]{2,}:[0-9a-f]{2,} \d+) +([^\n]+)$/m.match (contents,
```

oleavr (approved):
Whoah, great catch! Thanks a lot for digging into this :raised_hands:

PR #463, opened 2023-04-13
darwin: Make ThreadSuspendMonitor passthrough if called by Frida

mrmacete:
In this way we're sure this kind of crashes won't happen anymore:

```
Exception Type:  EXC_BAD_ACCESS (SIGBUS)
Exception Subtype: KERN_PROTECTION_FAILURE at 0x00000001bbfee294
Exception Codes: 0x0000000000000002, 0x00000001bbfee294
VM Region Info: 0x1bbfee294 is in 0x1bbfec000-0x1bbff4000;  bytes after start: 8852  bytes before end: 23915
      REGION TYPE                 START - END      [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      unused shlib __TEXT      1bbb01000-1bbfec000 [ 5036K] r-x/r-x SM=COW  ... this process
--->  __TEXT                   1bbfec000-1bbff4000 [   32K] r-x/rwx SM=COW  ..._kernel.dylib
      __TEXT                   1bbff4000-1bbff8000 [   16K] r-x/r-x SM=COW  ..._kernel.dylib
Exception Note:  EXC_CORPSE_NOTIFY
Termination Reason: SIGNAL 10 Bus error: 10
Terminating Process: exc handler [1185]

Triggered by Thread:  5

Thread 5 name:  gdbus
Thread 5 Crashed:
0   libsystem_kernel.dylib        	       0x1bbfee294 kevent + 8
1   ???                           	       0x108d11588 ???
2   ???                           	       0x108d10a3c ???
3   ???                           	       0x108d10c10 ???
4   ???                           	       0x108ccc690 ???
5   ???                           	       0x108d1f100 ???
6   libsystem_pthread.dylib       	       0x1dca8a338 _pthread_start + 116
7   libsystem_pthread.dylib       	       0x1dca88938 thread_start + 8
```

Which happened because:
- a Frida thread tried to execute some function while the page in which the function lived was writable
- `ThreadSuspendMonitor` prevented `thread_suspend` (called by the Interceptor) from working on that thread

this happened quite frequently at teardown time, when all Interceptor hooks get removed.

**Note**: ~~WIP because i'm going to test this change thoroughly~~ test succeeded

PR #466, opened 2023-04-18
agent: Dispose ThreadSuspendMonitor in its own transaction

mrmacete:
Since Interceptor relies on ThreadSuspendMonitor passthrough, disposing of it in its own transaction and after all the other components which use Interceptor at destruction time makes sure no frida thread will attempt executing code on non-executable pages at teardown time.

(This is the second part of https://github.com/frida/frida-core/pull/463)


oleavr (approved):
Thanks, great catch! 🤘

PR #469, opened 2023-04-25
darwin: Make spawn aware of prewarm

mrmacete:
Since iOS 15, Darwin has the "prewarm" feature for which most frequently used apps are being spawned ahead of time by `dasd` and kept suspended until the user attempts to launch them, in order to save some startup time.

This was in the way of Frida's spawn mechanism because:

- apps launched that way won't be going through launchd when expected, causing a "timeout was reached" error, or more complex race conditions
- Frida needs to spawn a fresh process to ensure early instrumentation

This change addresses these problems by:

- ignoring prewarm spawns from the launchd agent  (but they're still emitted as spawns if spawn gating is enabled)
- killing existing prewarmed apps before attempting to spawn them

The detection of prewarmed apps is done by checking the presence of the `ActivePrewarm=1` environment variable (libSystem makes the same assumption).

References:

- apple docs about prewarm: https://developer.apple.com/documentation/uikit/app_and_environment/responding_to_the_launch_of_your_app/about_the_app_launch_sequence#3894431
- libSystem: https://github.com/apple-oss-distributions/Libsystem/blob/1abeb883382de8a03e6fc0534c03115892382293/init.c#L522-L558

### DIFF
--- a/src/darwin/agent/launchd.js
+++ b/src/darwin/agent/launchd.js
@@ -51,13 +51,18 @@ applyJailbreakQuirks();
 
 Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dylib', '__posix_spawn'), {
   onEnter(args) {
+    const envp = args[4];
+    if (isPrewarmLaunch(envp)) {
+      return;
+    }
+
     const path = args[1].readUtf8String();
 
     let rawIdentifier;
     if (path === '/usr/libexec/xpcproxy') {
       rawIdentifier = args[3].add(pointerSize).readPointer().readUtf8String();
     } else {
-      rawIdentifier = tryParseXpcServiceName(args[4]);
+      rawIdentifier = tryParseXpcServiceName(envp);
       if (rawIdentifier === null)
         return;
     }
@@ -120,6 +125,26 @@ Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dyli
     }
   }
 });
+
+function isPrewarmLaunch(envp) {
+  if (envp.isNull())
+    return null;
+
+  let cur = envp;
+  while (true) {
+    const elementPtr = cur.readPointer();
+    if (elementPtr.isNull())
+      break;
+
+    const element = elementPtr.readUtf8String();
+    if (element.startsWith('ActivePrewarm'))
+      return true;
+
+    cur = cur.add(pointerSize);
+  }
+
+  return false;
+}
 
 function tryParseXpcServiceName(envp) {
   if (envp.isNull())
--- a/src/darwin/agent/launchd.js
+++ b/src/darwin/agent/launchd.js
@@ -52,9 +52,10 @@ applyJailbreakQuirks();
 Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dylib', '__posix_spawn'), {
   onEnter(args) {
     const envp = args[4];
-    if (isPrewarmLaunch(envp)) {
+    const prewarm = isPrewarmLaunch(envp);
+
+    if (prewarm && !gating)
       return;
-    }
 
     const path = args[1].readUtf8String();
 
@@ -70,7 +71,7 @@ Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dyli
     let identifier, event;
     if (rawIdentifier.startsWith('UIKitApplication:')) {
       identifier = rawIdentifier.substring(17, rawIdentifier.indexOf('['));
-      if (upcoming.has(identifier))
+      if (!prewarm && upcoming.has(identifier))
         event = 'launch:app';
       else if (gating)
         event = 'spawn';
--- a/src/darwin/agent/launchd.js
+++ b/src/darwin/agent/launchd.js
@@ -149,21 +149,14 @@ function parseStringv(p) {
 }
 
 function isPrewarmLaunch(env) {
-  for (const element of env) {
-    if (element.startsWith('ActivePrewarm='))
-      return true;
-  }
-
-  return false;
+  return env.some(candidate => candidate.startsWith('ActivePrewarm='));
 }
 
 function tryParseXpcServiceName(env) {
-  for (const element of env) {
-    if (element.startsWith('XPC_SERVICE_NAME='))
-      return element.substring(17);
-  }
-
-  return null;
+  const entry = env.find(candidate => candidate.startsWith('XPC_SERVICE_NAME='));
+  if (entry === undefined)
+    return null;
+  return entry.substring(17);
 }
 
 function applyJailbreakQuirks() {
--- a/src/darwin/agent/launchd.js
+++ b/src/darwin/agent/launchd.js
@@ -51,8 +51,8 @@ applyJailbreakQuirks();
 
 Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dylib', '__posix_spawn'), {
   onEnter(args) {
-    const envp = args[4];
-    const prewarm = isPrewarmLaunch(envp);
+    const env = parseStringv(args[4]);
+    const prewarm = isPrewarmLaunch(env);
 
     if (prewarm && !gating)
       return;
@@ -63,7 +63,7 @@ Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dyli
     if (path === '/usr/libexec/xpcproxy') {
       rawIdentifier = args[3].add(pointerSize).readPointer().readUtf8String();
     } else {
-      rawIdentifier = tryParseXpcServiceName(envp);
+      rawIdentifier = tryParseXpcServiceName(env);
       if (rawIdentifier === null)
         return;
     }
@@ -127,41 +127,40 @@ Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dyli
   }
 });
 
-function isPrewarmLaunch(envp) {
-  if (envp.isNull())
-    return null;
+function parseStringv(p) {
+  const strings = [];
 
-  let cur = envp;
+  if (p.isNull())
+    return [];
+
+  let cur = p;
   while (true) {
     const elementPtr = cur.readPointer();
     if (elementPtr.isNull())
       break;
 
     const element = elementPtr.readUtf8String();
-    if (element.startsWith('ActivePrewarm'))
-      return true;
+    strings.push(element);
 
     cur = cur.add(pointerSize);
+  }
+
+  return strings;
+}
+
+function isPrewarmLaunch(env) {
+  for (const element of env) {
+    if (element.startsWith('ActivePrewarm='))
+      return true;
   }
 
   return false;
 }
 
-function tryParseXpcServiceName(envp) {
-  if (envp.isNull())
-    return null;
-
-  let cur = envp;
-  while (true) {
-    const elementPtr = cur.readPointer();
-    if (elementPtr.isNull())
-      break;
-
-    const element = elementPtr.readUtf8String();
+function tryParseXpcServiceName(env) {
+  for (const element of env) {
     if (element.startsWith('XPC_SERVICE_NAME='))
       return element.substring(17);
-
-    cur = cur.add(pointerSize);
   }
 
   return null;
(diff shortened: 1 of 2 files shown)


src/darwin/frida-helper-backend-glue.m
```diff
@@ -1375,12 +1377,26 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
 
     timer = g_timer_new ();
 
-    while ([service pidForApplication:identifier] > 0 && g_timer_elapsed (timer, NULL) < kill_timeout)
+    while ((killed_pid = [service pidForApplication:identifier]) > 0 && g_timer_elapsed (timer, NULL) < kill_timeout)
```
oleavr, line 1380: Does this mean we always get stuck here waiting three seconds if a prewarmed app is being killed? If so it might be worth doing the check earlier, so we can just `SIGKILL` it right away and not have to wait.

(That reminds me it's unfortunate that we block the current thread's `MainContext` here -- frida-server is not able to do anything else while we're stuck here, when this function is called from `DarwinHelperBackend._kill_process()` and `DarwinHelperBackend._kill_application()`.)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
   return killed_pid;
 }
 
+static gboolean
+frida_has_active_prewarm (gint pid)
+{
+  gboolean result = false;
```
oleavr, line 1472:
```suggestion
  gboolean result = FALSE;
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+  int mib_argmax[] = { CTL_KERN, KERN_ARGMAX };
+  int mib_args[] = { CTL_KERN, KERN_PROCARGS2, 0 };
+  size_t size;
+  gint32 arg_max = 0, argc;
```
oleavr, line 1476: Think it would be clearer to initialize this explicitly on the line before it's used, i.e. just before the `sysctl()`.

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+
+  mib_args[2] = pid;
+  size = arg_max;
+
+  if (sysctl (mib_args, G_N_ELEMENTS (mib_args), buffer, &size, NULL, 0) != 0)
+    goto beach;
+  if (size < sizeof (argc))
+    goto beach;
+
+  argc = *(gint32 *)buffer;
```
oleavr, line 1498:
```suggestion
  argc = *(gint32 *) buffer;
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+beach:
+  g_free (buffer);
+
+  return result;
+}
+
+static guint8 *
+frida_skip_string (guint8 * cursor, guint8 * end)
+{
+  while (cursor < end && *cursor != 0)
```
oleavr, line 1529:
```suggestion
  while (cursor != end && *cursor != '\0')
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+    cursor++;
+  while (cursor < end && *cursor == 0)
+    cursor++;
```
oleavr, line 1532: Can there be more than one NUL that we want to skip? Wondering if this loop is needed and we could instead skip the one NUL we're already on. (That made the previous loop terminate.)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+
+  /* Skip executable name */
+  cursor = frida_skip_string (cursor, end);
+
+  /* Skip args */
+  while (argc-- != 0)
+    cursor = frida_skip_string (cursor, end);
+
+  /* Iterate environment */
+  while (cursor < end)
```
oleavr, line 1510:
```suggestion
  while (cursor != end)
```

(It should never go beyond unless there's a bug right?)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+  {
+    if (strstr ((char *)cursor, "ActivePrewarm=1"))
```
oleavr, line 1512:
```suggestion
    if (strstr ((char *) cursor, "ActivePrewarm=1") != NULL)
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+    {
+      result = true;
```
oleavr, line 1514:
```suggestion
      prewarm_active = TRUE;
```

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+  size_t size;
+  gint32 arg_max = 0, argc;
+  guint8 * buffer = NULL, * cursor, * end;
+
+  size = sizeof (arg_max);
+
+  if (sysctl (mib_argmax, G_N_ELEMENTS (mib_argmax), &arg_max, &size, NULL, 0) != 0)
+    goto beach;
+  if (arg_max < sizeof (argc))
+    goto beach;
```
oleavr, line 1484: Can't we trust the kernel here? (Or could a process be created with less?)

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1466,74 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
 
+static gboolean
+frida_has_active_prewarm (gint pid)
+{
+  gboolean result = false;
+  int mib_argmax[] = { CTL_KERN, KERN_ARGMAX };
+  int mib_args[] = { CTL_KERN, KERN_PROCARGS2, 0 };
+  size_t size;
+  gint32 arg_max = 0, argc;
+  guint8 * buffer = NULL, * cursor, * end;
```
oleavr, line 1477: Wonder if it would be more readable to move the assignment part down to just before where `size` is assigned. Since `buffer` is what this logical unit of the function is going to create, we could move it before `size` in the variable declarations, and assign it first, to kind of say "here we're about to compute *buffer*".

oleavr (changes requested):
Wow, cool!! :rocket:

src/darwin/agent/launchd.js
```diff
@@ -121,6 +127,26 @@ Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dyli
...
+
+    const element = elementPtr.readUtf8String();
+    if (element.startsWith('ActivePrewarm'))
+      return true;
+
+    cur = cur.add(pointerSize);
+  }
+
+  return false;
+}
```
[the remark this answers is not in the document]
mrmacete: yeah i was in doubt because in this way we would give up terminating the loop early but, as you said, most of the time we iterate it all at least once, so yeah parsing it properly once would be more readable and performance won't be worse and can be better in some cases.

on it

src/darwin/frida-helper-backend-glue.m
mrmacete: for not blocking the thread i wonder if this could be made async, by registering a SEND_DEAD callback on the about-to-bi-killed pid's task? anyways i think that belongs to a separate PR in a future time, because it's quite complex and can have weird edge cases which require proper testing IMHO.

oleavr: The easiest would probably be implementing it in Vala as an async method, where it uses a `TimeoutSource` to repeatedly check. But yeah that part is better left for a separate future PR.

> Does this mean we always get stuck here waiting three seconds if a prewarmed app is being killed?

My concern with this is that we always wait three seconds if the app has been prewarmed. Could we do the `has_active_prewarm()` before and go straight to `SIGKILL` in that case, or is this a bad idea?

src/darwin/agent/launchd.js
```diff
@@ -121,23 +127,42 @@ Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dyli
...
+  return strings;
+}
+
+function isPrewarmLaunch(env) {
+  for (const element of env) {
+    if (element.startsWith('ActivePrewarm='))
+      return true;
+  }
+
+  return false;
```
oleavr, line 157:
```suggestion
  return env.some(candidate => candidate.startsWith('ActivePrewarm='));
```

src/darwin/agent/launchd.js
```diff
@@ -121,23 +127,42 @@ Interceptor.attach(Module.getExportByName('/usr/lib/system/libsystem_kernel.dyli
...
+}
+
+function tryParseXpcServiceName(env) {
+  for (const element of env) {
+    if (element.startsWith('XPC_SERVICE_NAME='))
+      return element.substring(17);
+  }
+
   return null;
```
oleavr, line 166:
```suggestion
  const entry = env.find(candidate => candidate.startsWith('XPC_SERVICE_NAME='));
  if (entry === undefined)
    return null;
  return entry.substring(17);
```

src/darwin/frida-helper-backend-glue.m
oleavr: Oh wait, sorry, that's exactly what you did :blush:

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1462,73 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+  if (buffer == NULL)
+    goto beach;
+
+  mib_args[2] = pid;
+  size = arg_max;
+
+  if (sysctl (mib_args, G_N_ELEMENTS (mib_args), buffer, &size, NULL, 0) != 0)
+    goto beach;
+  if (size < sizeof (argc))
+    goto beach;
```
oleavr, line 1493: Is this possible?

src/darwin/frida-helper-backend-glue.m
```diff
@@ -1450,6 +1462,73 @@ static void frida_darwin_helper_backend_launch_using_sbs (NSString * identifier,
...
+  cursor = frida_skip_string (cursor, end);
+
+  /* Skip args */
+  while (argc-- != 0)
+    cursor = frida_skip_string (cursor, end);
+
+  /* Iterate environment */
+  while (cursor != end)
+  {
+    if (strstr ((char *)cursor, "ActivePrewarm=1") != NULL)
```
oleavr, line 1509:
```suggestion
    if (strstr ((char *) cursor, "ActivePrewarm=1") != NULL)
```

oleavr (changes requested):

src/darwin/frida-helper-backend-glue.m
mrmacete: i didn't observe this but looking at other code bases (chrome) they do it so i thought "better be on the safe side"

mrmacete: i can remove and retest

mrmacete: double checked in the kernel, and PROCARGS2 always have argc: https://github.com/apple-oss-distributions/xnu/blob/5c2921b07a2480ab43ec66f5b9e41cb872bc554f/bsd/kern/kern_sysctl.c#L1276

(`argc_yes` argument is passed as `1`)

src/darwin/frida-helper-backend-glue.m
mrmacete: yeah i initially implemented it differently because i thought the sysctl call would have been slow, instead is ridiculously fast 🤣

PR #471, opened 2023-05-03
fix memory leak in fallback kill code

tmm1:
introduced in https://github.com/frida/frida-core/commit/a081f8ff2211dbb0ae074b54cf9fb3b2f87c9034


oleavr:
Oops, nice catch! Thanks! :raised_hands:

PR #479, opened 2023-08-16
Fix spawn issues IOS 17 beta3

hsorbo:
Seems like new mangling of ProcessInfo

```
nm dyld.17b3 | grep ProcessInfo
00000001dd364b48 d __ZL12sProcessInfo
00000001a8619bb4 t __ZN5dyld423ExternallyViewableState23storeProcessInfoPointerEPP20dyld_all_image_infos
```

### DIFF
--- a/src/darwin/frida-helper-backend-glue.m
+++ b/src/darwin/frida-helper-backend-glue.m
@@ -2028,7 +2028,11 @@ _frida_darwin_helper_backend_prepare_spawn_instance_for_injection (FridaDarwinHe
     {
       instance->info_ptr_address = gum_darwin_module_resolve_symbol_address (dyld, "__ZN5dyld412gProcessInfoE");
       if (instance->info_ptr_address == 0)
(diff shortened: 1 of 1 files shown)


src/darwin/frida-helper-backend-glue.m
```diff
@@ -2028,7 +2028,11 @@ static void frida_darwin_helper_backend_launch_using_lsaw (NSString * identifier
     {
       instance->info_ptr_address = gum_darwin_module_resolve_symbol_address (dyld, "__ZN5dyld412gProcessInfoE");
       if (instance->info_ptr_address == 0)
-        goto dyld_probe_failed;
+      {
+        instance->info_ptr_address = gum_darwin_module_resolve_symbol_address (dyld, "__ZL12sProcessInfo");
```
oleavr, line 2032: Let's move this check first, for consistency with the existing logic where we check from newest to oldest.

PR #486, opened 2023-09-20
server: use sysroot for temporary files on rootless

fabianfreyer:
I have no clue about vala, so if there are things I should have done differently, I'd love to hear them :)


oleavr:
Thanks! 🙌  (You did great btw, I only applied a few nitpicks 👍)

PR #489, opened 2023-10-17
Fallback to "slower"  read_memory/write_memory on old Linux kernel

Pyraun:
## Issue

We faced some issues when launching a frida-server on a x86_64 box with an old Linux Kernel. The syscalls `process_vm_readv` and `process_vm_writev` exist but are not implemented.
Therefore, using those syscalls fail and the error ENOSYS is returned.

## Fix

We allow the linux backend to fallback to the slower implementation of  `read_memory`/`write_memory` when it faces this issue.
To avoid reaching this same issue at each call to `read_memory`/`write_memory`, we set the `process_vm_readv/process_vm_writev` function pointers to null to fallback directly to the "unoptimized" path.

### DIFF
--- a/src/linux/frida-helper-backend.vala
+++ b/src/linux/frida-helper-backend.vala
@@ -2141,10 +2141,9 @@ namespace Frida {
 				ssize_t res = process_vm_writev (pid, (Posix.iovector[]) &local, (Posix.iovector[]) &remote, 0);
 				if (res != -1)
 					return;
-				if (errno == Posix.ENOSYS) {
-					// The kernel does not implement the syscall, we avoid to use it by resetting the function pointer.
+				if (errno == Posix.ENOSYS)
 					process_vm_writev = null;
-				} else if (errno != Posix.EPERM && errno != Posix.EFAULT)
+				else if (errno != Posix.EPERM && errno != Posix.EFAULT)
 					throw new Error.NOT_SUPPORTED ("Unable to write to process memory: %s", strerror (errno));
 			}
 #endif
--- a/src/linux/frida-helper-backend.vala
+++ b/src/linux/frida-helper-backend.vala
@@ -2105,7 +2105,10 @@ namespace Frida {
 				ssize_t res = process_vm_readv (pid, (Posix.iovector[]) &local, (Posix.iovector[]) &remote, 0);
 				if (res != -1)
 					return result;
-				if (errno != Posix.EPERM)
+				if (errno == Posix.ENOSYS) {
+					// The kernel doesn't implement the syscall, we avoid to use it by resetting the function pointer.
+					process_vm_readv = null;
+				} else if (errno != Posix.EPERM)
 					throw new Error.NOT_SUPPORTED ("Unable to read from process memory: %s", strerror (errno));
 			}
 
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! Does this mean there are Linux kernels >= 3.2 that don't support these syscalls? I'm wondering if perhaps what's happening is that the kernel source code supports it, but the kernel was built with the feature disabled.

Pyraun:
It seems disabled or patched in my case which is a NAS device.

```bash
# uname -rsiopmv
Linux 4.14.22 #1 SMP Thu Oct 27 08:01:25 UTC 2022 Build-38 x86_64 unknown unknown GNU/Linux
```

```bash
> cat /proc/kallsyms | grep -i process_vm
ffffffffa80a6750 W compat_sys_process_vm_readv
ffffffffa80a6750 W compat_sys_process_vm_writev
> gdb /proc/kcore
(gdb) x/10i sys_process_vm_readv
No symbol table is loaded.  Use the "file" command.
(gdb) x/10i 0xffffffffa80a6750
   0xffffffffa80a6750:  mov    $0xffffffffffffffda,%rax
   0xffffffffa80a6757:  retq
   0xffffffffa80a6758:  nop
   0xffffffffa80a6759:  nop
   0xffffffffa80a675a:  nop
   0xffffffffa80a675b:  nop
   0xffffffffa80a675c:  nop
   0xffffffffa80a675d:  nop
   0xffffffffa80a675e:  nop
   0xffffffffa80a675f:  nop
(gdb) x/10i 0xffffffffa80a6750
   0xffffffffa80a6750:  mov    $0xffffffffffffffda,%rax
   0xffffffffa80a6757:  retq
   0xffffffffa80a6758:  nop
   0xffffffffa80a6759:  nop
   0xffffffffa80a675a:  nop
   0xffffffffa80a675b:  nop
   0xffffffffa80a675c:  nop
   0xffffffffa80a675d:  nop
   0xffffffffa80a675e:  nop
   0xffffffffa80a675f:  nop
(gdb) p/d 0xffffffffffffffda
$1 = -38
(gdb)
```

```bash
> errno 38
ENOSYS 38 Function not implemented
```

src/linux/frida-helper-backend.vala
```diff
@@ -2105,7 +2105,10 @@ namespace Frida {
 				ssize_t res = process_vm_readv (pid, (Posix.iovector[]) &local, (Posix.iovector[]) &remote, 0);
 				if (res != -1)
 					return result;
-				if (errno != Posix.EPERM)
+				if (errno == Posix.ENOSYS) {
+					// The kernel doesn't implement the syscall, we avoid to use it by resetting the function pointer.
+					process_vm_readv = null;
+				} else if (errno != Posix.EPERM)
```
oleavr, line 2111:
```suggestion
				if (errno == Posix.ENOSYS)
					process_vm_readv = null;
				else if (errno != Posix.EPERM)
```

src/linux/frida-helper-backend.vala
```diff
@@ -2139,7 +2142,10 @@ namespace Frida {
 				ssize_t res = process_vm_writev (pid, (Posix.iovector[]) &local, (Posix.iovector[]) &remote, 0);
 				if (res != -1)
 					return;
-				if (errno != Posix.EPERM && errno != Posix.EFAULT)
+				if (errno == Posix.ENOSYS) {
+					// The kernel does not implement the syscall, we avoid to use it by resetting the function pointer.
+					process_vm_writev = null;
+				} else if (errno != Posix.EPERM && errno != Posix.EFAULT)
```
oleavr, line 2148:
```suggestion
				if (errno == Posix.ENOSYS)
					process_vm_writev = null;
				else if (errno != Posix.EPERM && errno != Posix.EFAULT)
```

oleavr (approved):
Thanks! 🙌

PR #494, opened 2023-11-29
Added Error-Checking

47Cid:


src/api/merge.sh
```diff
@@ -6,6 +6,16 @@ build_dir="$3"
...
 
+if [ ! -e "$ar_binary" ]; then
+  echo "Error: ar_binary not found"
+  exit 1
+fi
+
+if [ ! -e "$libtool_binary" ]; then
+  echo "Error: libtool_binary not found"
+  exit 1
+fi
```
oleavr, line 17: Oops, sorry, spotted this too late and had to roll this back after merging: This will cause a regression since `libtool_binary` can be empty on systems where we don't use that tool.

PR #497, opened 2023-12-15
Add entitlements to allow read and execute cdhash

as0ler:
Adds  `com.apple.private.amfi.can-load-cdhash` and `com.apple.private.amfi.can-execute-cdhash` entitlements required to remap binaries in memory on iOS 16.


oleavr:
Thanks!

PR #498, opened 2024-01-09
gadget: Move TeardownRequirement parameter to Gum

mrmacete:
To allow Gum making better decisions taking into account its own promised lifetime (depends on https://github.com/frida/frida-gum/pull/768)


oleavr:
Sweet! Thanks! 🙌

PR #499, opened 2024-01-22
Fix Frida deb package for iOS 16.x

as0ler:
Fix Frida deb package to spawn frida-server when running on some specific iOS 16 versions (p.e. iOS 16.7.x)


as0ler:
LGTM!

PR #508, opened 2024-03-15
Add support for encoded mode

WorksButNotTested:
Support sending base64 between the TTY and the script to prevent extended ascii sequences getting converted into multibyte UTF-8 characters.

### DIFF
--- a/inject/inject.vala
+++ b/inject/inject.vala
@@ -414,7 +414,14 @@ namespace Frida.Inject {
 
 			buf[n] = 0;
 
-			script_runner.on_stdin ((string) buf);
+			if (script_runner.terminal_mode == TerminalMode.ENCODED)
+				/*
+				 * If we are in encoded mode, then send our data to
+				 * the target in base64 encoded form.
+				 */
+				script_runner.on_stdin (Base64.encode (buf[:n]));
+			else
+				script_runner.on_stdin ((string) buf);
 
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


inject/inject.vala
```diff
@@ -414,7 +414,14 @@ namespace Frida.Inject {
 
 			buf[n] = 0;
 
-			script_runner.on_stdin ((string) buf);
+			if (script_runner.terminal_mode == TerminalMode.ENCODED)
+				/*
+				 * If we are in encoded mode, then send our data to
+				 * the target in base64 encoded form.
+				 */
+				script_runner.on_stdin (Base64.encode (buf[:n]));
```
oleavr, line 422: No need to encode it, `post()` has a `data` argument  which supports binary data. It's passed to the message handler in the second argument. But we should improve the RPC API (built on top of `send()` + `recv()`) so it supports binary data in parameters and not only in the return value. The GumJS implementation can be found [here](https://github.com/frida/frida-gum/blob/main/bindings/gumjs/runtime/message-dispatcher.js). We could basically use a special marker, like `['frida:oob', offset, size]`, which specifies the offset into the `data` argument, and the `size` of the region.

oleavr (changes requested):

inject/inject.vala
WorksButNotTested: I think I had the same problem on output. Sending ansi control codes to the stdout. Is it possible for frida core to work as a dumb byte pump rather than dealing with strings?

oleavr: That's only a matter of returning an ArrayBuffer instead of a string.

WorksButNotTested: Can binary data only be sent/received by post and it's return value? Is there an ability to make an RPC to send bytes during the handling of the post? Or so they all have to be collated and sent by return?

WorksButNotTested:
Superseded by https://github.com/frida/frida-core/pull/525

PR #513, opened 2024-03-27
windows: Use RW/RX permissions for injection

yjugl:
This makes Frida injection compatible with more software. In particular, Mozilla Firefox rejects thread startup if the start address is RWX.

### DIFF
--- a/src/windows/frida-helper-backend-glue.c
+++ b/src/windows/frida-helper-backend-glue.c
@@ -305,10 +305,11 @@ frida_remote_worker_context_init (FridaRemoteWorkerContext * rwc, FridaInjection
   gpointer code;
   guint code_size;
   GumX86Writer cw;
-  const gsize data_alignment = 4;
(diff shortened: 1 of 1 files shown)


oleavr:
Yay, thanks for doing this! 🙌 Will land this as soon as I get the CI back in working order, so we can cut releases once again 😊

yjugl:
The issue was that the code below requires (one byte of) the `FridaRemoteWorkerContext` to be writable:

```c
void
frida_agent_main (const char * data, bool * stay_resident)
{
  (void) data;

  *stay_resident = true;

  append_to_log ('m');
}
```

I changed my patch to put the code and the `FridaRemoteWorkerContext` on different memory pages, so that they can have different permissions. This seemed like the easiest option. Alternatively we could probably make `FridaRemoteWorkerContext` readonly if we remove the `stay_resident` field. Then we would need to propagate that information in a different way, e.g. with the thread exit code or with a Windows event object.

oleavr:
Thanks! :fire:

PR #514, opened 2024-03-28
darwin: Take Interceptor and Cloak locks in TSM

mrmacete:
To extend the `ThreadSuspendMonitor` scope to prevent deadlock scenarios where threads holding the Cloak or the Interceptor lock get suspended.

depends on https://github.com/frida/frida-gum/pull/787


oleavr:
w00t! Thanks! :raised_hands:

PR #517, opened 2024-04-18
Fix frida-core on 32-bit systems

WorksButNotTested:


src/linux/frida-helper-backend.vala
```diff
@@ -1996,7 +1996,7 @@ namespace Frida {
 				get_regs (&saved_regs);
 				write_memory (target_address, original_code);
 
-				bool hit_breakpoint = saved_regs.program_counter == target_address ||
```
oleavr, line 1999: Hmm, I think we're trying to deal with a symptom here instead of the root cause. Somehow a 32-bit value must have been sign-extended to a `uint64` and ended up here.

I tried reproducing this locally on linux-x86, but couldn't. Do you have any tips on how to reproduce this on my end?

oleavr (changes requested):

src/linux/frida-helper-backend.vala
WorksButNotTested: I was using the stock 16.2.1 releases when working on `frida-cshell`, it seemed to fail reliably on both x86 and arm32. The project uses VScode `devcontainer` so should be pretty reproducible. It you clone [this](https://github.com/WorksButNotTested/frida-cshell/tree/master) and Ctrl-Shift-B to select one of the tasks you should be able to run it. It uses a patched version of frida though as you can see [here](https://github.com/WorksButNotTested/frida-cshell/blob/92b053dab0ee0a74512d2881f6a5114807b918d2/Dockerfile#L332).  You should just be able to copy the original release into the target environments and run it.

Also it is possible that rather than a sign extend issue, it could caused by dirty memory? e.g. a 32-bit value being written to only the low part of the 64-bit value, and then read back as 64-bits.

oleavr: Thanks!!

I'm so sorry, it's been so long since the last release that I totally forgot that I already fixed this issue post-16.2.1 back in March :sweat_smile: The fixes were: 7999a19eb43e064d264d32a755003cd78102a233 3cbefad1cdf940c06de2c8e91c8507ec2fdcb327

oleavr:
This was already fixed in 7999a19eb43e064d264d32a755003cd78102a233, and some related issues were fixed in 3cbefad1cdf940c06de2c8e91c8507ec2fdcb327. So sorry for the trouble!

WorksButNotTested:
> This was already fixed in 7999a19eb43e064d264d32a755003cd78102a233, and some other bugs just like it were fixed in 3cbefad1cdf940c06de2c8e91c8507ec2fdcb327. So sorry for the trouble!

Oh cool. No worries. So fixed in main, but not made it into a release just yet?

oleavr:
> > This was already fixed in 7999a19eb43e064d264d32a755003cd78102a233, and some other bugs just like it were fixed in 3cbefad1cdf940c06de2c8e91c8507ec2fdcb327. So sorry for the trouble!
> 
> Oh cool. No worries. So fixed in main, but not made it into a release just yet?

Yeah, will be part of the upcoming release.

PR #532, opened 2024-06-06
Add begin/end when building XPC arrays

hsorbo:


oleavr:
Thanks, great catch!! 🙌

PR #534, opened 2024-06-13
fruity: Add ProductType to query_system_parameters

as0ler:


src/fruity/fruity-host-session.vala
```diff
@@ -623,6 +623,7 @@ namespace Frida {
 				os["id"] = "ios";
 				os["name"] = properties.get_string ("ProductName");
 				os["version"] = properties.get_string ("ProductVersion");
+				os["type"] = properties.get_string ("ProductType");
```
oleavr, line 626: I'm thinking this belongs on the parent, as it's not a property of the operating system, but the system itself. Perhaps we should name it `model`, if that makes sense?

oleavr (changes requested):

src/fruity/fruity-host-session.vala
as0ler: Make sense. Updated!

oleavr (changes requested):
Thanks! 🤘  It seems `HWModelStr` is missing.  Could you give an example from `ideviceinfo` on a device where you're seeing such a value?  Wondering if it has a different name outside of MobileGestalt.

as0ler:
Adding HardwareModel into hardware model key.

PR #537, opened 2024-06-26
fruity: Add hwmodel, hwplatform and build information to query_system_parameters

as0ler:


oleavr:
Thanks! Rolled into #534.

PR #538, opened 2024-07-23
fruity: Fix adapter FirstUnicastAddress is NULL

xiofee:
Some virtual TUN adapter's FirstUnicastAddress is NULL, such as the Tunnel created by Clash, which will cause frida to crash when listing available devices.

SCR-20240723-kxgd-2


oleavr:
Thanks! 🙌

PR #539, opened 2024-07-25
Use maybe_extract_from_apk for Gadget files

gergesh:
On Android, when apps have `extractNativeLibs` set to `false`, the
configuration and potentially script files for frida-gadget are not
extracted and cannot be read using `FileUtils`.

### DIFF
--- a/lib/gadget/gadget.vala
+++ b/lib/gadget/gadget.vala
@@ -751,7 +751,7 @@ namespace Frida.Gadget {
 
 		string config_data;
 		try {
-			FileUtils.get_contents (config_path, out config_data);
+			load_asset_text (config_path, out config_data);
 		} catch (FileError e) {
 			if (e is FileError.NOENT)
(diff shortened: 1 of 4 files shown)


oleavr (changes requested):
Thanks! Sorry for the delay here.

I think it would be good to use the same library from the Vala code here, as the ElfModule API shouldn't be responsible for this.

gergesh:
I wasn't sure how to go about it without duplication of the code, care to point me in the right direction?

oleavr:
> I wasn't sure how to go about it without duplication of the code, care to point me in the right direction?

Just pushed a rough attempt at it. Sorry it took me so long to get to this. Haven't tested the code yet, but it should go something like this.

The idea is that once (minimal) minizip-ng Vala bindings are in place, the duplication needed is so minuscule that it's acceptable to duplicate the logic.

Issue #478, opened 2023-08-02
Under what circumstances would frida-server cause 100% CPU usage and become unresponsive?


lanyuer:
I started a frida-server, connected to it remotely, attached a process, and executed a script to call functions within the process. After running for some time, the frida-server becomes unresponsive. On my Pixel 4 phone, I noticed that the CPU is being exhausted, mainly by one process: unix:abstract=/frida-f4feb1bd-3f32-4282-bd which occupies 99.6% of the CPU.

I use strace to trace this process and sees that it keeps calling this piece of code:
`ppoll([{fd=3, events=POLLIN}, {fd=13, events=POLLIN}], 2, {tv_sec=0, tv_nsec=142951995325120880}, NULL, 0) = 0 (Timeout)
clock_gettime(CLOCK_MONOTONIC, {tv_sec=2278038456091393255, tv_nsec=4088476032}) = 0
ptrace(PTRACE_DETACH, 1452, NULL, SIG_0) = -1 ESRCH (No such process)
ptrace(PTRACE_INTERRUPT, 1452)          = -1 ESRCH (No such process)
ppoll([{fd=3, events=POLLIN}, {fd=13, events=POLLIN}], 2, {tv_sec=0, tv_nsec=142951995325120880}, NULL, 0) = 0 (Timeout)
clock_gettime(CLOCK_MONOTONIC, {tv_sec=2278596286443797735, tv_nsec=4088476032}) = 0
ptrace(PTRACE_DETACH, 1452, NULL, SIG_0) = -1 ESRCH (No such process)
ptrace(PTRACE_INTERRUPT, 1452)          = -1 ESRCH (No such process)
ppoll([{fd=3, events=POLLIN}, {fd=13, events=POLLIN}], 2, {tv_sec=0, tv_nsec=142951995325120880}, NULL, 0) = 0 (Timeout)
clock_gettime(CLOCK_MONOTONIC, {tv_sec=2279108590142864615, tv_nsec=4088476032}) = 0
ptrace(PTRACE_DETACH, 1452, NULL, SIG_0) = -1 ESRCH (No such^C{tv_sec=2295757257771695335, tv_nsec=4088476032}) = 0
ptrace(PTRACE_DETACH, 1452, NULL, SIG_0strace: Process 28541 detached`

The problem is that this process of 1452 no longer exists. Is there something wrong with the way I am using it?

asafe199:
I've been handling the issue here, I have a routine which runs my script for all day long.
[...]

PR #543, opened 2024-08-28
linux/frida-helper-backend: handle process gone

ajwerner:
This is an attempt to handle the situation where the process has gone away and we're shutting down. I've observed the system looping attempting to move the state out of ATTACHED. It's possible that the state machine here needs to be made more complex (ALREADY_ATTACHED) is misleading.

Informs #478

### DIFF
--- a/src/linux/frida-helper-backend.vala
+++ b/src/linux/frida-helper-backend.vala
@@ -1933,9 +1933,8 @@ namespace Frida {
 				close ();
 			} catch (Error e) {
 				// If the process is gone, then there's no point in retrying.
-				if (e is Error.PROCESS_NOT_FOUND) {
+				if (e is Error.PROCESS_NOT_FOUND)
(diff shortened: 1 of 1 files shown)


src/linux/frida-helper-backend.vala
```diff
@@ -1931,6 +1931,11 @@ namespace Frida {
 			try {
 				yield suspend (null);
 				close ();
+			} catch (Error e) {
+				// If the process is gone, then there's no point in retrying.
+				if (e is Error.PROCESS_NOT_FOUND) {
+					attach_state = ALREADY_ATTACHED;
+				}
```
oleavr, line 1938:
```suggestion
				if (e is Error.PROCESS_NOT_FOUND)
					attach_state = ALREADY_ATTACHED;
```

oleavr (approved):
Thanks, great catch! 🙌

PR #545, opened 2024-09-10
linux: Allocate stack for injector's remote calls

### LINKED ISSUE #544: provide mechanism for injector to use allocated stack
In some programming languages (Go, Java when using fibers) stacks are heap allocated and small. In this case, the way the injector hijacks a thread and assumes it has enough stack space is not valid. There should be some mode to have the injection logic allocate a stack to use for its remote calls.

ajwerner:
For some target programs it's not reasonable to assume that any hijacked thread has a large stack. For example, in Go, stacks are often small and are allocated on the heap. The injector bootstrap program uses kilobytes of stack. In order to side-step this problem, this patch enables an option to allocate an auxiliary stack for the remote call to use and, for the bootstrapper and loader, uses it.

The calls to mmap and munmap don't use much stack, so they are fine.

Fixes #544

### DIFF
--- a/src/linux/frida-helper-backend.vala
+++ b/src/linux/frida-helper-backend.vala
@@ -1017,8 +1017,9 @@ namespace Frida {
 				establish_connection (launch, spec, bres, agent_ctrl, fallback_address, cancellable);
 
 			uint64 loader_base = (uintptr) bres.context.allocation_base;
-
-			var call_builder = new RemoteCallBuilder (loader_base, saved_regs);
+			GPRegs regs = saved_regs;
+			regs.stack_pointer = bres.allocated_stack.stack_root;
+			var call_builder = new RemoteCallBuilder (loader_base, regs);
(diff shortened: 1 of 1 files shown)


ajwerner:
@oleavr I'm sure you'll have style comments and suggestions. I'd like to learn the right way to do things so I can more effectively help myself in the future. Let me know what you think of this when you find the time. Thanks!

ajwerner:
Okay, this patch relied on https://github.com/frida/frida-core/pull/550 to rebuild the bootstrapper images. I got rid of the build flags to enable this functionality from the original draft. I think this is a reasonable default.

oleavr:
Thanks! :raised_hands: Sorry it took me so long to get to this. Figured I'd make the final tweaks myself to not bother you with super-late feedback :sweat_smile:

ajwerner:
Thank you!

PR #546, opened 2024-09-17
linux: Improve injector's MUSL compatibility

luckycat889:
This should be harmless; it only affects musl platforms. The loader string could look different on different platforms


oleavr:
Thanks! :raised_hands: (And apologies for the delay!)

PR #547, opened 2024-09-17
Helpers; could we use triplets for the platform. Now, I don't really …

luckycat889:
Not sure if this could/should be accepted, as I might not understand the implications, but it'd make sense, since I have been experimenting with --host=i686-linux-musl and thus 'linux-x86-musl' as meson configuration


oleavr:
Thanks! :raised_hands:

PR #1122, opened 2024-11-03
Fix Compilation Errors due to Windows Encoding  Issues

JunGe-Y:
When I was compiling on windows, I got an error

`'utf-8' codec can't decode byte 0x90 in position 2:  invalid start byte"`
I looked at it because my windows system is gbk encoded, which is the encoding on many windows, and the frida-core compiler script is "utf-8" which caused this error, and when I modified it, the compiler worked fine.

### DIFF
--- a/compat/build.py
+++ b/compat/build.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 import argparse
 import base64
-import locale
 from collections import OrderedDict
 from dataclasses import dataclass, field
 import itertools
+import locale
 import os
 from pathlib import Path
 import pickle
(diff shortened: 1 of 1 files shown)


JunGe-Y:
There are many windows systems that are gbk encoded, but newer systems may already use utf-8 encoding, so I think getting it from the system might be a good choice

compat/build.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import annotations
 import argparse
 import base64
+import locale
```
oleavr, line 4: Imports should be sorted

oleavr (changes requested):
Thanks!

compat/build.py
JunGe-Y: Okay, I'm sorry. I've changed it

oleavr:
Thanks! :raised_hands:

PR #1131, opened 2025-01-10
Aarch64 be patches

WorksButNotTested:


oleavr (changes requested):
I suppose we also need to put in place automation for the two binary blobs, right?

WorksButNotTested:
> I suppose we also need to put in place automation for the two binary blobs, right?

Wouldn't do any harm if it's easy enough to do. But I don't imagine that changes in the loader and bootstrapper are that frequent?

PR #1134, opened 2025-01-21
Fix code swap region permissions

kaftejiman:


oleavr:
Thanks a lot! :raised_hands:

- I made some minor style tweaks to your libc matching patch, mainly introducing variables to avoid the need for comments.
- The suggested change to the code swap logic would require every such target to have at least one RWX region, which could be a bit much to ask and result in a regression. So I changed the logic there to skip memfd regions. For plain regions we don't need the W bit set, as ptrace() will help us out, but it appears it won't do that for memfd regions. Hopefully we won't have to add too many additional exceptions to this heuristic :)

PR #1139, opened 2025-03-11
Support building helpers for armbe8

WorksButNotTested:

PR #1140, opened 2025-03-11
Aarch64 be ilp32 patches

WorksButNotTested:
Note that this PR sits on top of:
* https://github.com/frida/frida-core/pull/1139
* https://github.com/frida/frida-core/pull/1131


src/linux/frida-helper-backend.vala
```diff
@@ -2031,7 +2031,11 @@ namespace Frida {
 				get_regs (&saved_regs);
 				write_memory (target_address, original_code);
 
-				bool hit_breakpoint = saved_regs.program_counter == target_address ||
+				/*
+				 * Note that on ILP32 platforms, these types are 64-bit, but in reality only
+				 * a 32-bit address space is in use.
+				 */
+				bool hit_breakpoint = (void *) saved_regs.program_counter == (void *) target_address ||
```
oleavr, line 2038: These casts shouldn't be needed unless junk bits have snuck into either side. So ISTM we should rather fix that if so? (This also doesn't change the second part of the condition, which would have the same issue.)

src/linux/helpers/Makefile
```diff
@@ -4,6 +4,8 @@ releng := $(top_srcdir)/releng
 BUILDDIR ?= $(top_srcdir)/build
 MESON ?= $(releng)/meson/meson.py
 
+meson := python3 $(FRIDA_MONOREPO)/releng/meson/meson.py
```
oleavr, line 7: This variable isn't used anywhere right?

src/linux/helpers/elf-parser.h
```diff
@@ -9,7 +9,7 @@
 #endif
 
 #ifndef __ELF_NATIVE_CLASS
-# if defined (__x86_64__) || defined (__aarch64__) || defined (__mips64)
+# if defined (__x86_64__) || ( defined (__aarch64__) && !defined (__ILP32__) ) || defined (__mips64)
```
oleavr, line 12:
```suggestion
# if defined (__x86_64__) || (defined (__aarch64__) && !defined (__ILP32__)) || defined (__mips64)
```

src/linux/helpers/nolibc-tweaks.patch
```diff
@@ -5,7 +5,7 @@ index f68baf8f3..18710dd30 100644
 @@ -181,6 +181,8 @@ struct sys_stat_struct {
  	_arg1;                                                                \
  })
- 
+
```
oleavr, line 8: Most of the changes to this file look like a text editor blindly took out all trailing whitespace. Please revert, so we can keep the history clean.

src/linux/helpers/nolibc-tweaks.patch
```diff
@@ -275,7 +287,7 @@ index fffdaf6ff..78eae91d0 100644
 +static void *memmove(void *dst, const void *src, size_t len)
  {
  	size_t dir, pos;
- 
+
```
oleavr, line 290: This file also has a bunch of whitespace changes that shouldn't be part of this.

src/linux/linjector.vala
```diff
@@ -84,9 +84,13 @@ namespace Frida {
 		public async uint inject_library_resource (uint pid, AgentDescriptor agent, string entrypoint, string data,
 				AgentFeatures features, Cancellable? cancellable) throws Error, IOError {
 			if (MemoryFileDescriptor.is_supported ()) {
+#if ARM64BEILP32
```
oleavr, line 87: Let's instead fix the underlying logic so this special-case isn't needed.

oleavr (changes requested):

src/linux/linjector.vala
```diff
@@ -85,7 +85,24 @@ namespace Frida {
...
+				string name;
+#if ARM64
+				/*
+				 * If our target is ARM, but our pointer size is not 8 bytes, then we must be building for ILP32.
+				 * We determine the bitness of our process by scanning for the high bits being set for pointer
+				 * types stored in the auxilliary vector. On ILP32, although our process is indeed 64-bit our
+				 * address space is constrained to 32-bits and hence we incorrectly determine the bitness, so
+				 * we override it here.
+				 */
+				if (sizeof (void *) == 4) {
```
oleavr, line 97: Could we move this quirk to the lowest level? (IIRC it's in Gum.)

src/linux/linjector.vala
```diff
@@ -85,7 +85,24 @@ namespace Frida {
...
+					name = "frida-agent-64.so";
+				} else {
+					name = agent.name_template.expand (arch_name);
+				}
+#else
+				name = agent.name_template.expand (arch_name);
+
```
oleavr, line 104:
```suggestion
```

oleavr (changes requested):

src/linux/helpers/nolibc-tweaks.patch
```diff
@@ -197,19 +197,31 @@ diff --git a/tools/include/nolibc/std.h b/tools/include/nolibc/std.h
...
++#endif
  typedef unsigned long        size_t;
  typedef   signed long       ssize_t;
- typedef unsigned long     uintptr_t;
- typedef   signed long      intptr_t;
+-typedef unsigned long     uintptr_t;
+-typedef   signed long      intptr_t;
++#if defined(__ILP32__)
++typedef unsigned int      uintptr_t;
++typedef   signed int      intptr_t;
```
oleavr, line 220: Doesn't seem visually aligned?

oleavr (changes requested):

PR #1144, opened 2025-03-17
fix iOS error "failed to open USB device: Input/Output Error"

jpstotz:
This PR fixes the issues:

* https://github.com/frida/frida/issues/3303
* https://github.com/frida/frida/issues/3278
* https://github.com/frida/frida/issues/3279
* https://github.com/frida/frida/issues/3283

As an Frida USB device with an empty `udid` should not occur with common iOS or Android devices I assume that it is safe to filter out such devices. 

@oleavr


oleavr:
Thanks! :raised_hands:

PR #1145, opened 2025-03-19
Support more than ~8 ADB-connected devices

philippmao:
If more than around 8 devices are connected over ADB, frida doesn't see any ADB connected devices anymore. This fixes the issue (tested for around 40 devices).


oleavr:
Thanks! :raised_hands:

PR #1150, opened 2025-04-07
channels: Break read loop on empty buffer

mrmacete:


oleavr:
Oops, my bad! 🤦‍♂️ Thanks! 🙌

PR #1154, opened 2025-04-10
Fix #1153 Update injector dyld init detection for iOS 18.4

### LINKED ISSUE #1153: iOS 18.4 _libdyld_initialize has changed so injector fails to detect as dyld4+
I ran into the issue described in https://github.com/frida/frida/issues/3424 when testing on iOS 18.4

Doing a quick look into this, the signature for _libdyld_initialize has changed, so the injector is failing to detect this as a dyld4 environment.

In iOS 18.3 (from IDA)
```
__text:00000001A8E9DF30 ; dyld4::APIs::_libdyld_initialize(dyld4::LibSystemHelpers const*)
__text:00000001A8E9DF30 __ZN5dyld44APIs19_libdyld_initializeEPKNS_16LibSystemHelpersE
__text:00000001A8E9DF30
```

in iOS 18.4 (from IDA)
```
; __int64 __fastcall dyld4::APIs::_libdyld_initialize(dyld4::APIs *__hidden this)
__ZN5dyld44APIs19_libdyld_initializeEv
```

Apparently the LibSystemHelpers pointer has been dropped in iOS 18.4.  So the code here is failing in this case since it's not also checking for the newer function:  https://github.com/frida/frida-core/blob/main/src/fruity/injector.vala#L828

pachoo:
added a few new dyld symbols there updated in iOS 18.4

### DIFF
--- a/src/fruity/injector.vala
+++ b/src/fruity/injector.vala
@@ -825,9 +825,9 @@ namespace Frida.Fruity.Injector {
 
 			yield restore_main_thread_state (cancellable);
 
-			uint64? libdyld_initialize = dyld_symbols["__ZN5dyld44APIs19_libdyld_initializeEPKNS_16LibSystemHelpersE"];
+			uint64? libdyld_initialize = dyld_symbols["__ZN5dyld44APIs19_libdyld_initializeEv"];
 			if (libdyld_initialize == null)
-				libdyld_initialize = dyld_symbols["__ZN5dyld44APIs19_libdyld_initializeEv"];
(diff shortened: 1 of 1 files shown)


oleavr (changes requested):
Thanks! :raised_hands: My only nitpick is to look for the newest symbol first. This is both for the reader to know that what we're looking for is progressively older, and having fewer lookup misses on newer OSes.

pachoo:
@oleavr appreciate the review, and agreed.  Reordered the lookups so that the iOS 18.4 symbols are first.

oleavr:
Thanks! :raised_hands:

PR #1157, opened 2025-04-11
network-stack: Reduce delay between writes

mrmacete:
By flagging as writable as soon as some space is available, without waiting for the low watermark.


oleavr:
w00t! Thanks! :raised_hands:

PR #1174, opened 2025-06-11
Be

WorksButNotTested:
* Fix the CI builds of labrats and helpers to only check the architecture that they are building.
* Use the prebuilt SDK for `armbe8`, `arm64be` and `arm64beilp32`
* Fix the disabling of a test which is supposed to be disabled in `armbe8` rather than `arm`


oleavr:
Thanks! :raised_hands:

PR #1175, opened 2025-06-12
Compiler/remove types

NSEcho:
This PR adds more symbol replacements that goes along with commit https://github.com/frida/frida-core/commit/375be6181a166a313d6d63d59c86525e44b578f0.

Additionally, this replaces weird `mingw_wgprintf` symbol that is nowhere defined, so it gets replaced with `mingw_vfprintf`. This is needed in order to allow CGO binaries to link with `libfrida-core.a` cross-compiled on Linux for Windows using mingw.


NSEcho:
Closing because the algo needs some tweaking

PR #1176, opened 2025-06-17
compiler: fix error reading shim files

lihe07:
Previously, embeddedShims can only reach source files with maximum depth of 2. Which will result in errors like:

open node_modules/@frida/assert/internal/util/comparisons.js: file does not exist

This update allows matching such files.

I have compiled and tested the modified frida compiler.


lihe07:
Also looking for a better & future-proof way of implementing that. 

Go embed currently does not support `**` glob, and does not have plans to include that.

Perhaps include all contents under `node_modules/@frida`? The total size is `622.7 KiB`.

oleavr:
Thanks! :raised_hands:

PR #1179, opened 2025-06-26
buffer: Fix max_length in read_fixed_string

mrmacete:
So it is constrained both within the requested size and the buffer's size.


oleavr:
Thanks! 🙌  Nice catch!

PR #1180, opened 2025-06-26
agent: hotfix NativeBridge to stop Exceptor from hooking libc

ptrstr:
Exceptor needs to hook `signal` and `sigaction`, but they are in `libc` leading to `gum_mprotect` aborting because it cannot change `libc`'s `r--`.

I tested on AVD with Android 14 and 15 and it prevents the crash observed in the upstream version when using `frida-server` or `frida-inject`.


oleavr:
Thanks! :raised_hands:

PR #1194, opened 2025-10-06
fruity: Unpair on InvalidHostID

mrmacete:
So the error doesn't become persistent, and the device can be paired again.


oleavr:
Thanks! 🙌

PR #1203, opened 2025-12-20
fruity: Throw CLOSED on tcp write in closed state

mrmacete:
In this way the stream stops being polled and an infinite loop is avoided.


oleavr:
Thanks! 🙌

PR #1205, opened 2026-01-12
compiler: rename Go conflictful symbols

NSEcho:
This PR adds more logic to remove troublesome Go symbols which is mostly the case when cross-compiling for Windows on Linux using mingw.

Will see now how test goes and if there is anything that needs to be fixed.

### DIFF
--- a/src/compiler/symbol-replacer/main.go
+++ b/src/compiler/symbol-replacer/main.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"strings"
 	"unicode"
 )
 


src/compiler/build-backend.py
```diff
@@ -85,8 +85,41 @@ def run(*args):
...
+
+            symbol_dest = priv_dir / "symbol-replacer"
+            symbol_dest.mkdir(exist_ok=True)
+            shutil.copy(base_dir / "symbol-replacer" / "main.go", symbol_dest)
+            shutil.copy(base_dir / "symbol-replacer" / "trie.go", symbol_dest)
+            shutil.copy(base_dir / "symbol-replacer" / "go.mod", symbol_dest)
+
+            symbol_replacer_sources = [str("symbol-replacer" / f.relative_to(symbol_dest)) for f in replacer_inputs if f.suffix == ".go"]
+
+            # when we are cross-compiling we want to unset GOOS and GOARCH
```
oleavr:
```suggestion
```

src/compiler/build-backend.py
```diff
@@ -85,8 +85,32 @@ def run(*args):
     )
 
     if mode == "c-archive":
-        symbols_to_scramble = detect_conflictful_symbols_in_archive(backend_a, config["nm"])
-        scramble_symbols_in_archive(backend_a, symbols_to_scramble, config["ranlib"])
+        if config.get("mingw") is None:
```
oleavr:
```suggestion
        if "mingw" not in config:
```

PR #1206, opened 2026-01-16
Support Segger J-Link

hsorbo:
Segger's JLinkGDBServer has these caveats:

- Insists on PROPER checksum
- Doesn't support vCont
- Doesn't do threads so no thread on read/write register.


oleavr:
Landed as part of the barebone-vsock-transport work. Thanks! :raised_hands:

PR #1211, opened 2026-02-16
fruity: Fix dbus starvation and backpressure

mrmacete:
this one depends on https://github.com/frida/libsoup/pull/1

### DIFF
--- a/lib/base/socket.vala
+++ b/lib/base/socket.vala
@@ -940,6 +940,9 @@ namespace Frida {
 					Memory.copy (buffer, recv_queue.data, n);
 					recv_queue.remove_range (0, (uint) n);
 
+					if (recv_queue.len == 0)
+						websocket.resume_input ();
+
 					update_pending_io ();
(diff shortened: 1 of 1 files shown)


oleavr:
Yay, nice! :raised_hands:  I guess this also needs a .vapi change in the Vala repo?

src/fruity/network-stack.vala
```diff
@@ -372,6 +378,18 @@ namespace Frida.Fruity {
 			return req.join ();
 		}
 
+		internal void perform_on_lwip_thread_nojoin (owned WorkFunc work) {
```
oleavr:
```suggestion
		internal void schedule_on_lwip_thread (owned WorkFunc work) {
```

Let's also move the meat of this method to a private method that this one and perform_on_lwip_thread() call. That helper returns the Request, which perform_on_lwip_thread() join()s, and this one just drops on the floor.

PR #1215, opened 2026-02-25
Fix Android Helper spawn crash on Infinix/Transsion devices

depreciating:
Fixes a NullPointerException that occurs when frida-server attempts to spawn an app on Infinix and other Transsion OS devices.

The Bug:
On these ROMs, the heavily modified ActivityThread implementation expects a fully initialized main looper. Using standard Looper.prepare() inside the Android Helper causes a crash inside android.os.Handler.<init> before the helper can inject.

The Fix:
Replaced Looper.prepare(); with android.os.Looper.prepareMainLooper(); inside Helper.java. I verified this manually by compiling the .dex and running it via app_process on a rooted Infinix device. The modified helper initializes cleanly and resolves the crash.


oleavr:
Thanks! :raised_hands:

PR #1225, opened 2026-04-10
dtx: Tolerate richer response shapes and name errors

ignuslabs:
Addressing https://github.com/frida/frida/issues/3719

## Problem

  Three DTX methods — `processIdentifierForBundleIdentifier:`,
  `isPidDebuggable:`, and
  `launchSuspendedProcessWithDevicePath:bundleIdentifier:environment:arguments:options:`
  — previously cast the reply directly to `NSNumber` and threw
  `Error.PROTOCOL ("Malformed response")` on anything else.

  On iOS 16 the remote service has been observed returning richer shapes
  (e.g. `NSDictionary` wrappers with conventional result keys, or error
  payloads such as `NSError` / dicts carrying `NSLocalizedDescription`),
  which the strict cast rejected. The result was an opaque
  `Malformed response` failure with no indication of which invocation
  failed or why — impossible to debug from a bug report.

  ## Fix

  - Introduce `parse_uint_response` / `parse_bool_response` helpers that
    accept `NSNumber`, parse numeric strings, or look up conventional
    result keys (`pid`, `processIdentifier`, `result`, `value`, …)
    inside an `NSDictionary`.
  - Detect error payloads via `try_extract_error_response_details`,
    which recognises `NSError`, string replies, and dict-wrapped errors
    (`NSError`, `NSLocalizedDescription`, `message`, …), and surface
    them as `Error.NOT_SUPPORTED` carrying the originating method name.
  - When no shape matches, `describe_response` stringifies the payload
    type and a truncated representation so future breakage is
    actionable instead of opaque.
  - Extend `DTXChannel` with a `PendingResponse` record that stashes the
    originating method name alongside each pending promise, so ERROR
    frames can report which invocation failed. The error-path
    formatting is consolidated through the new shared `format_nserror`
    helper.

  No behaviour change for the existing happy path: an `NSNumber` reply
  still resolves to the same value.

  ## Testing

  - Builds clean on macOS arm64 with `MACOS_CERTID=- make` in the
    top-level frida tree (full frida-gum + frida-core Vala + C pipeline).
  - `ninja src/libfrida-core.a` explicitly rebuilds the Vala compile
    pass over every frida-core `.vala` file, including the patched
    `dtx.vala`, which confirms the new helpers and `PendingResponse`
    inner class typecheck against the existing DTX code.
  - iOS-device-side verification of the new error path requires a
    reproducer on a physical device; this change is a strict superset
    of the prior cast behaviour so it is safe to merge even without
    such a reproducer in CI.

### DIFF
--- a/src/fruity/dtx.vala
+++ b/src/fruity/dtx.vala
@@ -362,11 +362,7 @@ namespace Frida.Fruity {
 				.append_object (new NSString (bundle_id));
 			var response = yield channel.invoke ("processIdentifierForBundleIdentifier:", args, cancellable);
 
-			NSNumber? pid = response as NSNumber;
-			if (pid == null)
-				throw new Error.PROTOCOL ("Malformed response");
-
-			return (uint) pid.integer;
+			return parse_uint_response ("processIdentifierForBundleIdentifier:", response);
 		}
 
 		public async bool is_pid_debuggable (uint pid, Cancellable? cancellable = null) throws Error, IOError {
@@ -374,11 +370,7 @@ namespace Frida.Fruity {
 				.append_object (new NSNumber.from_integer (pid));
 			var response = yield channel.invoke ("isPidDebuggable:", args, cancellable);
 
-			NSNumber? is_debuggable = response as NSNumber;
-			if (is_debuggable == null)
-				throw new Error.PROTOCOL ("Malformed response");
-
-			return is_debuggable.boolean;
+			return parse_bool_response ("isPidDebuggable:", response);
 		}
 
 		public async void start_observing_pid (uint pid, Cancellable? cancellable = null) throws Error, IOError {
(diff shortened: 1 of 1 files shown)


src/fruity/dtx.vala
```diff
@@ -511,6 +501,256 @@ namespace Frida.Fruity {
...
+				"PID",
+				"processIdentifier",
+				"ProcessIdentifier",
+				"processID",
+				"ProcessID",
+				"result",
+				"Result",
+				"value",
+				"Value",
+			};
```
oleavr, line 556: (Here and elsewhere:) This is a little too fuzzy/defensive for my taste. Can we make this more explicit? Each specific caller should know what shapes exist for the specific call, and use helper functions to avoid duplication.

oleavr (changes requested):
Thanks! 🙌  Some further refinement is needed:

Issue #1227, opened 2026-04-15
Feature request for syscall-trace


IPMegladon:
Hey I love the new eBPF syscall tracer and am currently building a TUI tool that makes use of it.

Would it be feasible to add `include-syscall` functionality to the `syscall-trace` service?
While `exclude-syscalls` works fine if you only set this once, if you want to add and remove which calls are traced at runtime a `include-syscall` function would be very useful to re-include a function after filtering it out. 
Currently I am doing the filtering client side which works but given the sheer amount of syscalls with some apps this isn't ideal. 

I don't mind taking a stab at implementing this myself but I still have a very limited understanding of how this codebase works.

oleavr:
Glad to hear that ☺️ Yes, that's definitely a welcome addition 👌🏻

PR #1228, opened 2026-04-15
(feat) Add support for (re)including syscalls with SyscallTracer

IPMegladon:
#1227 add `include-syscall` functionality to the `syscall-trace` service.
Have tested against android, but not yet iOS.
Ended up being a smaller change than I expected.


oleavr:
Thanks! 🙌

PR #1229, opened 2026-04-17
fruity: Detach lldb only if no attached gadgets

mrmacete:
before: when multiple gadget entries relied on the same lldb session, the first to detach also detached lldb
after: only the last detached gadget entry actually detaches lldb


oleavr:
Thanks, great catch! 🙌

PR #1234, opened 2026-05-14
android: fix package name when spawned activity has android:process

pakbaz-dev:
If an activity declares android:process, the process forked from zygote has a name like "com.example.app:foo" instead of "com.example.app". The zymbiote sends process name "com.example.app:foo" in its hello, but spawn_request holds the package name "com.example.app". Since the names don't match, it keeps waiting for the spawn to complete until timeout is reached.

We strip the colon suffix from the process name before matching, since the part after colon is not part of the package name. Also rename package_name to process_name to reflect what it actually contains.


oleavr:
Thanks, great catch! 🙌

PR #1241, opened 2026-06-11
Document the public API

oleavr:
## Summary

Adds API reference documentation to frida-core's public surface, mirroring the effort in frida-gum. Docs are written as idiomatic **valadoc** comments in the `.vala` sources and flow into the generated `Frida-1.0.gir`, so they render through the same gi-docgen pipeline as frida-gum on the website.

## How it works

`valac --gir` strips doc comments, so this teaches `src/api/generate.py` — which already merges the core + base GIRs into `Frida-1.0.gir` — to pick the documentation up itself:

- Extracts `/** … */` blocks from the `.vala` sources (the top-level files plus `lib/base/**`), associating each with the declaration it precedes.
- Converts valadoc markup to the Markdown gi-docgen expects (`{{{ }}}` → fenced code, `''bold''`, `//italic//`, etc.).
- Resolves `{@link Foo.bar}` into real gi-docgen cross-references — `[class@Frida.Session]`, `[method@Frida.Device.attach]`, `[property@Frida.Device:id]`, `[signal@Frida.Device::output]`.
- Injects `<doc>` into the merged GIR, including per-`@param` and `@return` docs, with async `_sync`/`_finish` twins, property getter/setter methods, and Vala default constructors all handled.

No build-graph change — the docs live in the source and are emitted by the existing API-generation step.

## Coverage

**76/76 types** and **596/597 members** documented and verified in the generated GIR (the one exception is a `protected` internal helper). This spans the whole public API: `DeviceManager`/`Device`/`Session`/`Script`, spawn/attach/enumeration, the option and list types, errors and enums, the portal/control services, networking/auth, RPC, the `Compiler`, and the `PackageManager`.

## Commits

- `api: Inject Vala doc comments into the GIR` — the pipeline.
- `core: Document the public API` — `frida.vala`.
- `core: Document base enums and option types`.
- `core: Document portal, peer, and web types`.
- `core: Document compiler, packages, and services`.

## Notes

Verified end-to-end by running the modified `generate.py` against prebuilt core/base GIRs (no full frida-core build was done locally); the first CI build will confirm against freshly generated GIRs.

### DIFF
--- a/src/frida.vala
+++ b/src/frida.vala
@@ -11,8 +11,17 @@ namespace Frida {
 	public extern void version (out uint major, out uint minor, out uint micro, out uint nano);
 	public extern unowned string version_string ();
 
+	/**
+	 * Selects which main-loop runtime Frida integrates with.
+	 */
 	public enum Runtime {
+		/**
+		 * Use GLib's own main loop.
+		 */
 		GLIB,
+		/**
+		 * Integrate with a foreign runtime that hosts GLib.
+		 */
 		OTHER;
 
 		public static Runtime from_nick (string nick) throws Error {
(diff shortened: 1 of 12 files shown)


src/frida.vala
```diff
@@ -468,32 +581,96 @@ namespace Frida {
...
+		 * @return the device
+		 */
 		public new Device get (int index) {
 			return items.get (index);
 		}
 	}
 
+	/**
+	 * Represents a device that Frida can interact with, such as the local
+	 * system, a USB-connected phone, or a remote agent.
```
oleavr:
```suggestion
	 * system, a USB-connected phone, or a remote frida-server.
```

PR #1244, opened 2026-06-21
linux: Fix 'Unable to pick a payload base' on Android 10

wave-sky:
Spawning processes on Android 10 fails with a remote connection error due to Android's XOM mechanism. The mechanism restricts `libstagefright.so` to execute-only access, causing the read permission check in `do_prepare_zymbiote_injection` to fail.

Fix this by:
- Removing the read permission check for target libraries.
- Expanding the list of eligible media shared libraries.
- Forcing read-execute permissions to ensure payload execution.

### DIFF
--- a/src/linux/linux-host-session.vala
+++ b/src/linux/linux-host-session.vala
@@ -1687,9 +1687,10 @@ namespace Frida {
 			} catch (GLib.Error e) {
 				throw_api_error (e);
 			}
-		}
+		}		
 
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! Landed as 1b540b96c2635c55e0b0daf59bdde64167fbb748 (squashed with a small whitespace/formatting tidy-up). 🎉

PR #1247, opened 2026-07-15
Protect segments before applying fixups

hsorbo:
The upload injector maps the agent module r-x, streams in its segments, applies threaded and chained fixups, and only afterwards restores each segment's protection.

The fixup passes write the rebased and authenticated pointers directly into the module (e.g. into __DATA_CONST). Because the data segments are still mapped r-x at that point, on arm64e the first authenticated-rebase store faults with KERN_PROTECTION_FAILURE and the injection is aborted with "Invocation of 0x... crashed".

Apply each segment's protection before running the fixup passes, so the data segments are writable when the pointers land. __TEXT is not fixed up and stays r-x, so no rwx mapping is created, and the WRITE command keeps populating the module through its own temporary writable mapping regardless of ordering.

### DIFF
--- a/src/fruity/injector.vala
+++ b/src/fruity/injector.vala
@@ -507,6 +507,19 @@ namespace Frida.Fruity.Injector {
 					yield output.write_all_async (bytes.get_data (), io_priority, cancellable, out bytes_written);
 				}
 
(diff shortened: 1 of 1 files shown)


oleavr:
Thanks! 🙌

PR #1251, opened 2026-08-07
fix: allow admin, local service and system accounts access to the com…

jamiechapmanbrn:
I make a custom agent w/ frida embedded in order to rewrite network calls on the fly over a proxy.

In testing I found that for some processes mysteriously frida would fail to attach
eg.
`error code 4: Process with pid 6284 either refused to load frida-agent, or terminated during injection`

In the end we tracked down that the issue was happening when we were attaching to processes that were running as local service and they were having trouble accessing the pipe.

This change adds some extra sddl to add some privileged groups to the sddl that allow that attach to occur.


oleavr:
Yay, nice catch! Thanks! :raised_hands:

PR #1253, opened 2026-08-19
Allow wildcards at the beginning or end of a pattern

Xoffio:
Related to frida/frida-gum#1143 and frida/frida-gum#1144


oleavr:
Thanks! 🙌

PR #1255, opened 2026-09-14
Update default values and disguise service name

luyeguigen:
Changed default address and port values, modified thread name and identifier to disguise as a system service.


oleavr:
Thanks for the PR, and for taking the time to put this together.

I'm going to decline this one, though. The changes here — renaming the thread/service, changing the default address and port, and adding the LUOYE_INJECT_SCRIPT fallback — are all about disguising Frida from RASP-style detection. That's a direction I've deliberately kept out of upstream.

The reasoning: baking evasion into Frida just kicks off a never-ending cat-and-mouse game. The moment a particular disguise ships in the default build, it becomes the new fingerprint for RASP vendors to detect, and we're back where we started — except now everyone's binary looks identical and is trivially matched.

I'd much rather leave this as an exercise for the reader. Frida is easy to patch, and when everyone applies their own tweaks — their own names, ports, and layouts — the result is a huge space of permutations that's genuinely hard to enumerate. That diversity is far more effective than any single evasion we could ship, and it keeps upstream neutral and predictable for the vast majority of users who aren't fighting RASP at all.

So please keep this as a local patch — it's a perfectly good one to carry downstream. Thanks again for contributing, and I hope the reasoning makes sense.

PR #1257, opened 2026-09-19
Preserve suspension across policy softening

jiska2342:
frida-server softens a target by attaching and detaching with `ptrace`. Attaching resumes a suspended process. A process that Frida spawns/spawn gates is started suspended so that it can be instrumented before it runs any of its own code. Since the attach has already resumed it, Frida's check for a suspended target reports false, so it skips the step that waits for dyld to finish and injects into a process whose libraries are not loaded yet. This fails with `module not found at "/usr/lib/libSystem.B.dylib"`, and sometimes `unable to read from task memory`.

This patch makes softening leave the suspension state as is. It takes its own suspension before attaching and drops it afterwards only if the attach did not consume it, so the target stays suspended until frida resumes it.

Tested on an SRD with iOS 27.0 beta: a spawn followed by an attach goes from succeeding 1 in 8 to 8 in 8, and spawn gating works.


oleavr:
Thanks! 🙌

